PR #1, opened 2012-12-16
Add timeout option for socks proxies

nferch:


schwabe:
I think my proposed patch (http://mid.gmane.org/1444919912-4432-1-git-send-email-arne@rfc2549.org) implements this in a better way.

cron2:
@schwabe is this already solved in some way, or will it be solved as soon as The Big Timeout Patch is merged?

schwabe:
@cron2 No the timeout patch removes the socks-proxy-timeout and integrates it into the big timeout

dsommers:
As I understand it, commit f2134b7bea37df15756c599b94f16d4bffafbbd6 seems to be resolving this issue.  So closing this one.

PR #2, opened 2013-02-17
Fix MSVC build

jait:
Trivial fix to make openvpn buildable with MSVC.


cron2:
2.3.x builds with MSVC right now, git master doesn't (for other reasons), so this one can really be closed.  thanks.

PR #3, opened 2013-04-18
add obfuscate to openvpn connections ( very important for the users in china)

sadan9:
Chinese Government use GFW ( Great FireWall Of China .about it please look this:http://en.wikipedia.org/wiki/Great_Firewall_of_China) to block lot of website ,like Facebook , Twitter ,Youtube....

some  chinese people use PPTP , L2TP ,OPENVPN , stunnel to bypassing GFW.
but after Dec 2012, GFW can block the connections of Openvpn. it's over looks like over Layer 7 because GFW can block the connections with any ports of TCP or UDP protocols.
some people think openvpn have feature codes in TLS handshake.

this change add obfuscate for openvpn connections to against the GFW.
there are tow parameter add to openvpn config file:
obfs-salt [secret]
 -------this parameter  is secret for obfuscate . 
obfs-padlen [num] 
------- this parameter  optional. num=1~255. openvpn will add random data in the end of packages

### DIFF
--- a/src/openvpn/options.c
+++ b/src/openvpn/options.c
@@ -62,6 +62,10 @@
 
 #include "memdbg.h"
 
+extern char* _socket_obfs_salt;
+extern int _socket_obfs_salt_len;
+extern int _socket_obfs_padlen;
+
 const char title_string[] =
   PACKAGE_STRING
   " " TARGET_ALIAS
@@ -6820,6 +6824,19 @@ add_option (struct options *options,
       options->persist_mode = 1;
     }
 #endif
+  else if (streq (p[0], "obfs-salt") && p[1])
+    {
+      VERIFY_PERMISSION (OPT_P_GENERAL);
+      _socket_obfs_salt = p[1];
+      _socket_obfs_salt_len = strlen(_socket_obfs_salt);
+    }
+  else if (streq (p[0], "obfs-padlen") && p[1])
+    {
+      VERIFY_PERMISSION (OPT_P_GENERAL);
+      _socket_obfs_padlen = atoi(p[1]);
+      if (_socket_obfs_padlen < 0)
+       msg(M_ERR, "--obfs-padlen must be positive");
+    }
   else
     {
       if (file)
--- a/src/openvpn/socket.c
+++ b/src/openvpn/socket.c
@@ -41,6 +41,10 @@
 
 #include "memdbg.h"
 
+const char* _socket_obfs_salt = NULL;
+int _socket_obfs_salt_len = 0;
+int _socket_obfs_padlen = 0;
+
 const int proto_overhead[] = { /* indexed by PROTO_x */
   0,
   IPv4_UDP_HEADER_SIZE, /* IPv4 */
@@ -51,6 +55,49 @@ const int proto_overhead[] = { /* indexed by PROTO_x */
   IPv6_TCP_HEADER_SIZE,
   IPv6_TCP_HEADER_SIZE,
 };
+
+/**
+ * @return int The length of the random string that should be padding to the packet
+ */
+int obfs_buffer(const struct buffer* buf, const void* rand, int randlen, int maxpadlen) {
+  unsigned char md[SHA_DIGEST_LENGTH];
+  unsigned char iv[randlen + _socket_obfs_salt_len + SHA_DIGEST_LENGTH];
+  unsigned char *c;
+  int i, len, pad_len = 0;
+
+  if (maxpadlen > 255)
+    maxpadlen = 255;
+
+  /* key_1 = SHA1(rand + obfs_salt) */
+  /* pad_len = Low _rand_pad_level_ bits of (unsigned char)MD5(rand + obfs_salt)[0] */
+  memcpy(iv, rand, randlen);
+  memcpy(iv + randlen, _socket_obfs_salt, _socket_obfs_salt_len);
+
+  /* Caculate length of padding string */
+  ASSERT(SHA_DIGEST_LENGTH >= MD5_DIGEST_LENGTH);
+  MD5(iv, randlen + _socket_obfs_salt_len, md);        /* SHA_DIGEST_LENGTH is bigger than MD5_DIGEST_LENGTH, it's safe here */
+  if (maxpadlen <= 0)
+    pad_len = 0;
+  else
+    pad_len = md[0] % (maxpadlen + 1);
+
+  /* Obsfucation data */
+  len = BLEN(buf);
+  SHA1(iv, randlen + _socket_obfs_salt_len, md);
+  for (i = 0, c = BPTR(buf); i < len; i++, c++)
+  {
+    *c ^= md[i % SHA_DIGEST_LENGTH];
+
+    /* Regenerate obsfuction key: key_n+1 = SHA1(key_n) */
+    if (i % SHA_DIGEST_LENGTH == SHA_DIGEST_LENGTH - 1)
+    {
+      memcpy(iv, md, SHA_DIGEST_LENGTH);
+      SHA1(iv, SHA_DIGEST_LENGTH, md);
+    }
+  }
+
+  return pad_len;
+}
 
 /*
  * Convert sockflags/getaddr_flags into getaddr_flags
--- a/src/openvpn/socket.h
+++ b/src/openvpn/socket.h
@@ -36,6 +36,10 @@
 #include "socks.h"
 #include "misc.h"
 
+extern int _socket_obfs_salt_len;
+extern int _socket_obfs_padlen;
+int obfs_buffer(const struct buffer* buf, const void* rand, int randlen, int rand_pad_level);
+
 /*
  * OpenVPN's default port number as assigned by IANA.
  */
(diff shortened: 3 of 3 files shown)


lipnitsk:
@ngharo Was there a decision made at the developer meeting?

cron2:
Hi,

On Mon, Oct 12, 2015 at 03:41:07PM -0700, Ilya Lipnitskiy wrote:
> @ngharo Was there a decision made at the developer meeting?

No, we ran out of time.

gert

-- 
                                                           //www.muc.de/~gert/

naivefun:
+1

alimakki:
Can't this better be done with something like obfsproxy?
Perhaps making it easier to configure OpenVPN with obfs4 would be a more ideal approach.

sadan9:
some device can't run obfs4. 
ext. openvpn for iOS…

chipitsine:
https://github.com/OpenVPN/openvpn/pull/22#issuecomment-145233027

mattock:
The OpenVPN development team has talked about adding obfuscation to OpenVPN many times in the past. We think obfuscation should be handled outside of OpenVPN by software that is designed for the job (e.g. obfsproxy). As mentioned above, it is a cat-and-mouse game, and we don't want to play that game. I believe obfsproxy does not need to be running locally (on the phone), so even if certain restricted platforms are unable to run it, there are workarounds. From what I've heard, commercial OpenVPN providers also tend to bundle obfsproxy instead of bundling obfuscation patches with OpenVPN.

I'll close this pull request soon unless somebody from the developer team objects.

cron2:
Hi,

On Tue, May 17, 2016 at 11:52:21PM -0700, Samuli Seppänen wrote:
> I'll close this pull request soon unless somebody from the developer team objects.

We're not adding obfuscation code to OpenVPN - because, as you have mentioned,
it's an arm's race, so whenever we deploy something, the people that we want
to obfuscate from can see our commits and deploy countermeasures...

What I think would make sense is to add a plugin API for "packet mangling"
(so OpenVPN would create, encrypt and sign its outgoing packets and hand
it off to a plugin for obfuscation before sending, and incoming packets 
would be de-obfuscated by the plugin before signature check / decryption).

I'm not going to implement this API any time soon, but I'm willing to 
discuss and review, and potentially merge - provided the API is sane, and
does not add significant overhead to un-obfuscated packet flows.

gert
-- 
                                                           //www.muc.de/~gert/

dsommers:
Also see pull request #7 for similar arguments to why not use such static mangling/obfuscation methods into OpenVPN.

johnx7271:
In reply to @cron2
An simple and effective way to obfuscate
Since openvpn is built on ssl, a simple way to obfuscate is to make it look like a https webserver until someone tell a spell (a secret code, like 芝麻开门) . Imagine you stand before a painting on a wall, before you say "芝麻开门", it remain a painting, no mater whatever you do or you say, but after  you say a spell, a door is suddenly wide-open for you, and then you can do all the openvpn stuff.
To implement this, just re-route all the traffic which do not have a certain header, to a webserver (whatever, internal or external), and for the client to get this header, it must tell a spell.

cron2:
Hi,

On Sat, Feb 24, 2018 at 10:57:22AM -0800, johnx7271 wrote:
> In reply to @cron2
> An simple and effective way to obfuscate
> Since openvpn is built on ssl, a simple way to obfuscate is to make it look like a https webserver until someone tell a spell (a secret code, like ????????????) . Imagine you stand before a painting on a wall, before you say "????????????", it remain a painting, no mater whatever you do or you say, but after  you say a spell, a door is suddenly wide-open for you, and then you can do all the openvpn stuff.
> To implement this, just re-route all the traffic which do not have a certain header, to a webserver (whatever, internal or external), and for the client to get this header, it must tell a spell.

This is already in place.  You can run openvpn on tcp/443 and all normal
http connects get proxied to a http(s) server, and only openvpn client
handshakes get handled as "openvpn".

This is not going to help with oppresive regimes, though - they are not
stupid, and can read source code, so they know the secret handshake.

To repeat the official answer by the developers: we are not going to
implement obfuscation in the mainline code, but we *are* looking into
providing a plug-in API so everyone can have their own obfuscation 
module (= it's their arm's race, not ours).

gert

-- 
now what should I write here...

schwabe:
@johnx7271 VPN traffic without obfuscation really easy to identify just by traffic pattern analaysis. I think you really underestimate your adversary here.

dsommers:
What @schwabe says.  For example, an https site doesn't have a constant stream of a relative similar packet size - which in VPN cases would typically be below 2KB (or probably even closer to 1500 bytes).  So proper obfuscation need to account for bundling up bigger packets, which increases VPN tunnel latency. And it needs to have at least some periods with a longer or shorter packet send/receive delay.  All this hits the general VPN tunnel performance in different ways.

And then comes the initial SSL/TLS handshake, which you need to comply to to have a proper SSL/TLS protocol.  Here the certificates of server and client are transferred across the link unencrypted - where you can easily extract the certificate subjects and the reference to the signing CA.  This means you can't even easily (or at least for a long time) even use fake "netflix.com" subject to disguise it as Netflix streaming - because it would be easy to spot it is a fake certificate as the signing CA would not match the official netflix.com certificates.

Yes, you might get around this for some time - but just until the DPI firewalls learns these tweaks and improves their detection algorithms.

Regardless, this obfuscation (even through a pure SSL/TLS connection) needs to happen outside the core OpenVPN.  We are in the process of getting a plug-in API in place for OpenVPN so such obfuscation can be handled outside the core OpenVPN.  This way DPI firewalls will have it more difficult as the obfuscation is not necessarily a "standard" obfuscation.  It depends entirely on the plug-ins being deployed.  We will most likely provide some simple sample plug-ins for obfuscation.  But for larger scale deployments, it would be advisable to use a quite different obfuscation algorithm.

johnx7271:
I agree that vpn traffic has abnormal statistical attributes for gfw detecting, and site certificate is also an issue.  But isn't ssl the best foundation on which to build obfuscating?  1. ssl is secure, so it is very easy to hide something in it, and is very difficult for the middle man to have a full monitor of it. 2. ssl is widely used.
I am not familiar with vpn and openvpn implementation, I suppose that, openvpn operates on top of ssl, so all the openvpn traffic is completely hidden from middle man, for example, it can be designed such that, the server check for a spell in client request (or handshake), if it's correct, then it goes on to vpn processing flow, else it simply re-route to a web server. The spell is wrapped in ssl data payload, so probably it is not easy to use plun-in to implement these.

dsommers:
The OpenVPN wire protocol contains of a data channel and a control channel.  The type of channel is indicated in on of the first bytes of the packet being sent over the wire.  The control channel deploys the TLS protocol, where the configuration between client and server is exchanged as well as a key derivation process.  This is also where the certificate/PKI handling is covered.  The result of the key derivation process via the control channel is used as a symmetric encryption key which is used on the data channel.  And it is the data channel which carries the tunnelled network data.

SSL/TLS itself is fairly robust and safe.  But remember that when deploying that on-top of OpenVPN itself, you get two layers of encryption.  This carries a noticable performance cost as well. And it is essentially pointless in regards to the tunnel security itself, as OpenVPN already uses the TLS protocol.  But the packets reveals themself easily due to the special approach of encapsulating SSL/TLS packets for the control channel.  The additional SSL/TLS layer only needs to disguise the first, say, 8 bytes, of a packet.  And if used together with ``--tls-crypt`` (instead of ``--tls-auth``), the control channel is encrypted - including the certificate handshake.  An obfuscation method taking this approach will have a far better performance than adding encryption on top of everything in the packet.

And what you describe with separating and re-routing https traffic to a web server ...  @cron2 already correctly pointed out that this "magic" is already present.  Have a look at ``--port-share`` in the [man page](https://community.openvpn.net/openvpn/wiki/Openvpn24ManPage).

Now is the time to stop beating a dead horse.

johnx7271:
"deploying that on-top of OpenVPN"? so you designed openvpn this way? I thought it should be openvpn over tsl/ssl.  Then there would be no need for double encryption, and it can be made completely indistinguishable from tsl/ssl.

Anyway thank you.

selvanair:
FWIW, I detect some misunderstanding here. Some are talking about obfuscation within openvpn and providing plugin or other support for it. I think that is useless for bypassing GFW or anything of that level of sophistication and resources.

The other suggestion is  to use openvpn over a TLS connection. I guess, no one responded because (i) that's not what this thread originally was about (ii) it doesn't require any support from openvpn --- e.g., just use stunnel as the TLS layer. That alone is easily detected unless TLS client auth is used and traffic to a real web server is multiplexed.  A multiplexer like [sslh](http://www.rutschle.net/sslh) or portshare within openvpn itself may help.

In a very limited testing I participated,  openvpn through stunnel appeared to work well through GFW. But long running servers with sustained traffic will very likely alert GFW through statistical analysis.

Recap: Obfuscation is best handled outside OpenVPN and is more effectively discussed on forums dedicated to it.

chipitsine:
@selvanair - what do you think, it seems to make sense to add obfsproxy support to openvpn-gui (like privatetunnel and airvpn already do) ?

selvanair:
My first reaction would be no, but if any one submits a good quality patch that is easy to maintain, we could consider it. If you want to discuss this further, please start a thread in the GUI. This one has been beaten to death..

johnx7271:
@selvanair，my "openvpn over a TLS connection" means different than yours, I mean:
 openvpn should be (vpn specific stuff and plus some camouflage stuff) + tls, 
(Guess that means a re-writing of openvpn, a suggestion too late)
you mean: 
(the current entire openvpn) over(tunneled thru) tls.

dsommers:
@johnx7271 said:
> "deploying that on-top of OpenVPN"? so you designed openvpn this way? I thought it should be openvpn over tsl/ssl. Then there would be no need for double encryption, and it can be made completely indistinguishable from tsl/ssl.

Well, it has to be this way to get the best performance.  SSL/TLS depends a bit too much on asymmetric encryption which is by far slower than symmetric encryption.  I also believe there is a larger packet overhead with "proper" SSL/TLS connections than pure symmetric encryption.  And the traffic we pass over the control channel ("proper" SSL/TLS) is fairly minimal to the amount of network data being transported (over the symmetrically encrypted data channel).

A few words on the impact the overhead in each packet results in. For example, users switching from ``--auth SHA1`` (default [1]) to ``--auth SHA256``  does report a measurable performance loss, as this change makes each OpenVPN packet 24 bytes longer.  And users who deployed ``--auth SHA512`` are really experiencing a more sluggish connection, as that grows each packet with 64 bytes per packet compared with SHA256.  So each byte of overhead counts.

Another detail of this design.  SSL/TLS is strictly designed for TCP connections, with reliability aspects (packet drops, resends, duplicate packets check, etc) being handled by the TCP stack.  VPNs have the best performance with UDP (TCP over TCP is known challenge with VPNs), so the approach used by OpenVPN enables SSL/TLS over UDP - as the OpenVPN encapsulation adds the needed information in UDP packets to handle the reliability stuff  built into the TCP stack.

Now you can argue that DTLS handles this with UDP.  But DTLS was published in April 2006.  The very first initial OpenVPN release arrived in May 2001 and the first version deploying a control channel was released in October 2002.  And the OpenVPN wire protocol is mostly backwards compatible back to at least OpenVPN 2.0 - as long as you don't use OpenVPN features not available in such old clients.  (Not that we recommend using such old clients)

So that is both the technical and historic background for the current design.

[1] Usage of SHA1 in the ``--auth SHA1`` context is considered safe, as it is used with HMAC.  HMAC-SHA1 is not currently known to be broken.

dsommers:
@chipitsine said:
> @selvanair - what do you think, it seems to make sense to add obfsproxy support to openvpn-gui (like privatetunnel and airvpn already do) ?

It is truly sad to not see these VPN service providers interacting with the community to resolve this issue in a far more transparent way.

Once we get the plug-in interface involved, their hackish approach will no longer be needed.  You just need to compile the obfuscation plug-in for the platforms you support and deploy that.  Whatever this plug-in does can be whatever they implement it to do, so it could still carry some competition edge even though the core OpenVPN got a more native support for obfuscation.

dsommers:
I will now lock this discussion, I think we've reached as long as we can come.

PR #4, opened 2013-07-06
tls_ctx_load_ca: Improve certificate error messages

kdienes:
If a CA certificate file includes intermediate certificates, and any
of them fail to verify, the current code will file with "Cannot load
CA certificate file".  Instead, generate a more specific error message
identifying the specific sub-certificate(s) which did not validate.

I make no claim that I really understand the OpenVPN code, but here's the scenario that led to this patch:

I was using the "pile of certificates" handed to me by StartSSL, and one of them was expired, but unnecessary to the verification chain.  So when I tried to verify using openssl, it worked fine, but OpenVPN complained with a not-very-helpful "Cannot load CA certificate file."  The following patch was what I ended up doing to figure out what was going wrong.

I have the suspicion that there are more complex procedures for submitting a patch that I don't know about, but I figured I'd start like this to see if the change was interesting enough to you for me to pursue it further.

### DIFF
--- a/src/openvpn/ssl_openssl.c
+++ b/src/openvpn/ssl_openssl.c
@@ -693,7 +693,7 @@ tls_ctx_load_ca (struct tls_root_ctx *ctx, const char *ca_file,
   X509_STORE *store = NULL;
   X509_NAME *xn = NULL;
   BIO *in = NULL;
-  int i, added = 0;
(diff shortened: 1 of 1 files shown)


cron2:
since it has been merged, the pull request needs closing :-)

PR #5, opened 2013-07-15
Support for username-only auth file.

mludvig:
Make OpenVPN read the username from the auth file
parameter of --auth-user-pass and prompt for a
password if it's not in the file.

Rationale: Prior to this change OpenVPN either
required both username and password present in the
auth file or prompted for both on the console.
Unlike passwords usernames usually don't change and
can therefore be "hardcoded" in the config.

Signed-off-by: Michal Ludvig <mludvig@logix.net.nz>


cron2:
Took a while, but has been merged now (thanks!)

commit 6e9373c84639382c16d9eb8f1f78f60079bb89df (master)
commit 76b82f0e9d272a703624b102bd2e2b38d9947afb (release/2.3)
Author: Michal Ludvig <mludvig@logix.net.nz>
Date:   Sun Oct 11 10:44:20 2015 +0200

    Support for username-only auth file.

PR #7, opened 2013-08-09
Bypass DPI

russlank:
Hi,

As some ISP(s) rely on Deep Packet Inspection (DPI) to identify various VPN traffic, and to block them, I have added small feature, called “bypass-dpi”. This new feature mainly gives OpenVPN the ability to masquerade the protocol by sending arbitrary, meaningless, initial packet that confuses the DPI mechanism, and allows consequent valid connection's packets to pass through, without being identified as OpenVPN traffic.

As the DPI appliances establish context for each newly established connection passing through, depending on examining known protocols’ signatures, at very early stage of connection establishment, sending such a packet makes DPI blind, and makes it classifying the protocol as unknown.

Currently implemented method works with specific OpenVPN configuration:
- UDP protocol to be used for the connection.
- TLS-AUTH to be specified; this will allow the server to drop the initial invalid packet, which means that the support for the feature is needed on the client side only.

I have modified the code to include this feature, and which can be enabled using “bypass-dpi” option in the configuration or at command-line.

Frankly, I made the change with the minimum effort. I think it could be implemented in better, more consistent way, by someone who has better understanding of the code.

Russlan

### DIFF
--- a/config-msvc.h
+++ b/config-msvc.h
@@ -19,6 +19,9 @@
 #define ENABLE_PORT_SHARE 1
 #define ENABLE_SOCKS 1
 #define ENABLE_SSL 1
+/* Added by RusslanK: BEGIN */
+#define ENABLE_BYPASS_DPI 1
+/* Added by RusslanK: END */
 
 #define HAVE_ERRNO_H 1
 #define HAVE_FCNTL_H 1
--- a/src/openvpn/forward.c
+++ b/src/openvpn/forward.c
@@ -1141,6 +1141,7 @@ process_outgoing_link (struct context *c)
 	    /* Send packet */
 
       /* Added by RusslanK: BEGIN */
+#ifdef ENABLE_BYPASS_DPI
       if (c->options.bypass_dpi)
         if (!c->c2.first_data_sent)
         {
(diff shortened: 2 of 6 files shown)


shel3over:
what about this patch ?
https://github.com/clayface/openvpn_xorpatch/blob/master/openvpn_xor.patch

dsommers:
The issue with both the XOR patch and this patch is exactly what has been mentioned here, DPI analysers gets better and better and these approaches - even though they can work very well now - will at some point stop working again.

In general we recommend using obfsproxy by the Tor project.  This can work together with OpenVPN out of the box using SOCKS5 mode and --sock-proxy in the OpenVPN config.  On the server side you also need to run obfsproxy listening to an alternative port and then tell it to proxy inbound connections to your OpenVPN server port.

The advantage with obfsproxy is that it doesn't have a purely static obfuscation - and the Tor project have the possibility to enhance the obfs protocol as well.  Considering that the Tor project have far more users in the same situation, they should be able to have both the know-how and the interest in fixing issues with the obfsproxy far quicker.

The downside of this approach is that you currently must use OpenVPN in TCP mode.

For more information, see https://www.torproject.org/projects/obfsproxy.html.en

PR #8, opened 2013-08-19
netsh workaround when interface is utf-8 named

chipitsine:
please, have a look at  https://community.openvpn.net/openvpn/attachment/ticket/318


cron2:
The original bug has been fixed long ago, but since then we've changed it all again - and are now using interface index numbers when calling netsh.exe on Vista and up, so any funnies in interfaces names should be solved for good now.

PR #12, opened 2014-01-22
Apple Mac OS X Keychain support

csdexter:
(rebased on HEAD, as requested)

Hello,

This is a resurrection of the original patch submitted by Bran Raderman brian@irregularexpression.org four years ago. I took the liberty of porting the patch to the current OpenVPN source style and layout, while at the same time adapting it to follow in the footsteps of the Microsoft CryptoAPI support as closely as possible.

This has been tested to work fine on Mac OS X 10.9.1 using a binary compiled using the XCode 4.x command line tools and GNU autotools installed using brew.

Please have a look,
Kind regards,
Radu - Eosif Mihailescu

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -330,6 +330,7 @@ case "$host" in
 		have_tap_header="yes"
 		dnl some Mac OS X tendering (we use vararg macros...)
 		CPPFLAGS="$CPPFLAGS -no-cpp-precomp"
+		MACOSX=yes
 		;;
 	*-mingw*)
 		AC_DEFINE([TARGET_WIN32], [1], [Are we running WIN32?])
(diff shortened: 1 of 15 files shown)


cron2:
I think this has been implemented in the  meantime, in

commit 39e3d336d4eeab847a3395ddeb430e0a9ca387b9
Author: Vasily Kulikov <segoon@openwall.com>
Date:   Wed Feb 25 19:07:18 2015 +0300

    Mac OS X Keychain management client
    
    This patch adds support for using certificates stored in the Mac OSX
    Keychain to authenticate with the OpenVPN server.

csdexter:
Yes it has, one year after I refactored the 4 year old prior art :-) Also, the current implementation is bad design: it takes away the management connection so if you're on Mac OS X and want to do anything else with the management interface besides read certificates from Keychain, you cannot.

The management interface doesn't support more than one client at a time and nobody wrote a multiplexing proxy for it yet -- so my implementation was designed to mimic the CryptoAPI support for MS Windows specifically so that the management interface remains available.

I'm not arguing against the OpenVPN developers' decision over which implementation to include, just wanted to note the implementation subtleties here for posterity.

PR #14, opened 2014-03-10
Fix SOCKSv5 method selection

Yawning:
So, RFC 1928 doesn't say anything about the METHODS field in the Method
Selection message being ordered in terms of preference or anything, and
the server is free to pick any of the METHODS offered by the client.

Always sending a Method Selection message with NO AUTHENTICATION REQUIRED
and USERNAME/PASSWORD set is broken on two fronts:

 * If the OpenVPN client can't handle the server picking USERNAME/PASSWORD
   due to the credentials being missing, it shouldn't offer it to the server.

 * If the OpenVPN client has credentials, then it should always attempt to
   authenticate.  This is a security product.  "You can misconfigure it and
   it will work" is not acceptable.  Setting a username/password when the
   SOCKS server doesn't require/support that as an option is the user not
   configuring it correctly, and should be treated as such.

### DIFF
--- a/src/openvpn/socks.c
+++ b/src/openvpn/socks.c
@@ -255,6 +255,13 @@ socks_handshake (struct socks_proxy_info *p,
       return false;
     }
 
+  /* validate that the auth method returned is the one sent */
+  if (buf[1] != method_sel[2])
+    {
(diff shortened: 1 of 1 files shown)


cron2:
since this has long been merged, we can just close it :-)

PR #15, opened 2014-04-10
Fix typo in build script to use LDFLAGS

kangsterizer:
just in case someone actually passes them as environment, might as well work :)


dsommers:
Patch applied, see commit a95358af543b9106f4ef481e4556d1d03459d058commit

PR #19, opened 2014-12-13
Export IPv6 pool options to scripts

choppsv1:
- Export the IPv6 analogous options to IPv6 ifconfig_pool_*

This is needed for things like adding ND proxy, or ip6tables access entries in client connect scripts.


dsommers:
This seems to be covered by commit a8f8b9267183c3cfc065f344d61effe6c55c3da6  ... -> closing

PR #20, opened 2015-03-03
mixed bag: Make openvpn management kill command support killmsg

cuckoohello:
It is not easy to find cid for using client-kill command, so make kill command support killmsg is very useful for stopping remote client with AUTH_FAILED.


schwabe:
Can you send that patch to openvpn-devel mailing list?

ordex:
Closing as it is old and does not apply anymore.

PR #21, opened 2015-03-25
src/openvpn/buffer.c

wangxiaowei0303:
1. improve rm_trailing_chars()


cron2:
Indeed, our code is not very efficient, calling strlen() every time - OTOH, this usually only removes one or two characters (being called from chomp() etc.), so it does not really make a difference.

Your code doesn't follow our coding style, though, so we would not merge it - *and*, we do not do pull requests.  So if you want it included, please change the bracket style to be in line with the existing coding style, and send it by mail (git-send-email) to openvpn-devel@lists.sourceforge.net

thanks.

PR #22, opened 2015-04-12
systemd: Improve handling of runtime directory.

eliasp:
- Don't hardcode `/var/run` but use `%t` instead to determine the
  runtime base directory depending on the environment

- Define `RuntimeDirectory=` and `RuntimeDirectoryMode=` to create e.g.
  `/run/openvpn` on demand when it's non-existent

### DIFF
--- a/distro/systemd/openvpn-client@.service
+++ b/distro/systemd/openvpn-client@.service
@@ -8,11 +8,10 @@ Documentation=https://community.openvpn.net/openvpn/wiki/HOWTO
 
 [Service]
 PrivateTmp=true
-Type=forking
 RuntimeDirectory=openvpn
 RuntimeDirectoryMode=0710
-PIDFile=%t/openvpn/client_%i.pid
-ExecStart=/usr/sbin/openvpn --cd /etc/openvpn/client --config %i.conf --daemon --writepid %t/openvpn/client_%i.pid
+WorkingDirectory=/etc/openvpn/client
+ExecStart=/usr/sbin/openvpn --config %i.conf --suppress-timestamps
 CapabilityBoundingSet=CAP_IPC_LOCK CAP_NET_ADMIN CAP_NET_BIND_SERVICE CAP_NET_RAW CAP_SETGID CAP_SETUID CAP_SYS_CHROOT CAP_DAC_READ_SEARCH
 LimitNPROC=10
 DeviceAllow=/dev/null rw
--- a/distro/systemd/openvpn-client@.service
+++ b/distro/systemd/openvpn-client@.service
@@ -9,8 +9,10 @@ Documentation=https://community.openvpn.net/openvpn/wiki/HOWTO
 [Service]
 PrivateTmp=true
 Type=forking
-PIDFile=/var/run/openvpn/client_%i.pid
-ExecStart=/usr/sbin/openvpn --cd /etc/openvpn/client --config %i.conf --daemon --writepid /var/run/openvpn/client_%i.pid
+RuntimeDirectory=openvpn
+RuntimeDirectoryMode=0710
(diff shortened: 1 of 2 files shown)


eliasp:
I've added a few more improvements. Copy of the commit msg:
```
commit e24993427a2a927443ec2948ea5c333246384ce9
Author: Elias Probst <mail@eliasprobst.eu>
Date:   Sun Sep 27 21:56:21 2015 +0200

Further improvements to the systemd service units

Use more of systemd's builtin functionality instead of trying to deal
with this in the service itself:

- don't double-fork (`--daemon`) but keep the process instead attached
  to its parent.
  This obsoletes the need for complex/possibly error-prone process
  tracking through a PID file.
- don't print timestamps to systemd's journal. The systemd journal has a
  dedicated field per message which provides a neutral (UTC-based) value
  which is converted to the suitable format when viewing the journal
  (`--suppress-timestamps`).
- Change the working directory through systemd (instead of `--cd …`)
```

syzzer:
Thanks for giving your improvements back to the project. However, OpenVPN does not accept pull requests. See http://community.openvpn.net/openvpn/wiki/DeveloperDocumentation. (Tl;dr: Use `git format-patch` to create patches, then email them to openvpn-devel@lists.sourceforge.net using git send-email.)

mattock:
Now that we allow use of PRs for code review - how does this code look?

dsommers:
I'll put this on my todo list as well, seems reasonable.  And I wonder if this also can resolve the suggested improvement in PR #28.

dsommers:
I've finally had some time to look at this.  I want to have this sorted out for the next 2.4 alpha release.

I tested this on one of my production boxes with Scientific Linux 7.2 (RHEL7.2 clone).  And something odd happens when using --chroot in the openvpn config.

Options error: Temporary directory (--tmp-dir) fails with '/var/chroot/openvpn/tmp': Permission denied

If I run the command line systemd would use manually, it works as expected.  If I remove the --chroot option, it works as expected.  I've tried to disable PrivateTmp,  but that didn't help.  Debugging now why it fails.

dsommers:
This is interesting.  When enabling chroot **and** the temp directory inside the chroot is **not** owned by root, the CAP_DAC_OVERRIDE is needed when doing initial file checks. 
If the temp dir is owned by root, CAP_DAC_READ_SEARCH is enough.

Okay, I'll pull in most of these changes, and a few other minor changes and submit this to the mailing list.

eliasp:
The `chroot` option of OpenVPN seems redundant to [systemd.exec](https://www.freedesktop.org/software/systemd/man/systemd.exec.html)'s generic capabilities to isolate the capabilities and filesystem permissions of services mostly through namespacing (e.g. [ReadWritePaths=](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#ReadWritePaths=), [PrivateNetwork=](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#PrivateNetwork=) etc.) and a plain chroot is most likely by far not as secure as proper namespacing (not having looked at the internal implementation details yet to really judge what's done there).

dsommers:
That's a fair point.  But we need to consider all the configurations out in the wild.  So for now, we need to ensure --chroot works well and then we can start thinking about hardening this further.

The --chroot implementation is fairly naive.  It calls chroot() and chdir("/") and then it drops uid/gid.  A good starting point looking at this would be init.c:947 do_uid_gid_chroot()

dsommers:
Sent to mailing list.
Message-Id: [<1476996158-5390-1-git-send-email-davids@openvpn.net>](https://www.mail-archive.com/openvpn-devel@lists.sourceforge.net/msg12764.html)

PR #24, opened 2015-06-14
__func__ is not supported on Visual Studio 2010. Replace it with a macro

ltfish:
:-)

(After double-checking, I found out that I was using the compiler in VS 2010, not 2013)


syzzer:
This was fixed in a simpler way in 9884e20810bda737c7708ff587e09cc0bb8475c7. This pull request can be closed.

dsommers:
Closing.

PR #26, opened 2015-06-19
connection management: Implement session timeout parameter.

zelenkovsky:
In multiple case it is important to limit user connection time. At the moment it is only
possible by using management console and writing complicated script to follow connection status.

There is an easier way to do that now. "session" parameter allows to define timeout in seconds
after which user will be disconnected. Session timeout can be declared globally or per user in
ccd file. Session is started at the moment user connects to VPN and will be automatically terminated
when timer reaches the zero.

This feature can be used together with Radiusplugin for OpenVPN in order to support Session-Timeout attribute.

Patch for Radiusplugin is coming as well.

### DIFF
--- a/src/openvpn/forward-inline.h
+++ b/src/openvpn/forward-inline.h
@@ -118,6 +118,19 @@ check_inactivity_timeout (struct context *c)
     check_inactivity_timeout_dowork (c);
 }
 
+/*
+ * Should we exit due to session timeout?
+ */
+static inline void
+check_session_timeout (struct context *c)
+{
+  void check_session_timeout_dowork (struct context *c);
+
+  if (c->options.session_timeout
+      && event_timeout_trigger (&c->c2.session_interval, &c->c2.timeval, ETT_DEFAULT))
+    check_session_timeout_dowork (c);
+}
+
 #if P2MP
 
 static inline void
--- a/src/openvpn/forward.c
+++ b/src/openvpn/forward.c
@@ -322,6 +322,16 @@ check_inactivity_timeout_dowork (struct context *c)
   register_signal (c, SIGTERM, "inactive");
 }
 
+/*
+ * Should we exit due to session timeout?
+ */
+void
+check_session_timeout_dowork (struct context *c)
+{
+  msg (M_INFO, "Session timeout (--session), exiting");
+  register_signal (c, SIGTERM, "session");
+}
+
 #if P2MP
 
 void
@@ -527,6 +537,11 @@ process_coarse_timers (struct context *c)
   if (c->sig->signal_received)
     return;
 
+  /* possibly exit due to --session */
+  check_session_timeout (c);
+  if (c->sig->signal_received)
+    return;
+  
   /* restart if ping not received */
   check_ping_restart (c);
   if (c->sig->signal_received)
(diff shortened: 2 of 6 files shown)


schwabe:
I think my proposed patch http://mid.gmane.org/1444919912-4432-1-git-send-email-arne@rfc2549.org should solve this in a better way.

cron2:
sorry for being such slackers regarding patches.  This is actually interesting work and might help me solve a problem I have at a customer site :-) - so, reviving this, and will look into it.

If you have time, rebasing the patch would be welcome, to ensure it still applies to latest master (options.c tends to change a lot, and quite often breaks patches adding new options)

zelenkovsky:
No problem, I'll try to rebase my work. But to make it work, you need to install separate plugin for Radius authentication. I had to patch it as well.

cron2:
Hi,

On Tue, Jun 07, 2016 at 02:33:36PM -0700, Dmitry Zelenkovsky wrote:
> No problem, I'll try to rebase my work. But to make it work, you need to install separate plugin for Radius authentication. I had to patch it as well.

Shouldn't it work by passing the parameter from a ccd/ file or a 
client-connect script?  Or by just setting it in the main config?

(All my users have the same rule, which is "nobody connects longer 
than 12h", so a global setting would be fine)

gert
-- 
                                                           //www.muc.de/~gert/

zelenkovsky:
Yes, it works in a following way:

1) Radius plugin handles authentication and writes ccd file with parameters received from radius server.
2) One of parameters is Session-Time out and it is written into ccd as well as others.
3) After authentication is done OpenVPN reads ccd file and assigns parameters to the session.
4) It was implemented that session time out is read from ccd and after this number of seconds OpenVPN disconnects.
5) After connection is broken, Radius plugin deletes ccd file (optional)

So session timeout attribute can be assigned statically for the user by preparing ccd file. Then user will be always disconnected after pre-defined number of seconds. Or better option - to use Radius authentication plugin which will read the value from Radius server.

galtenberg:
Would love to see this get in 👍

ordex:
Any of the involved parties would like to revive this PR and rebase it against latest master?

ordex:
Moreover, any idea why the ping-timeout wouldn't be enough?

cron2:
Hi,

On Sat, Sep 17, 2022 at 12:38:54PM -0700, Antonio Quartulli wrote:
> Moreover, any idea why the ping-timeout wouldn't be enough?

ping-timeout is active on inactivity.  session-timeout is something
like "you have paid for your VPN subscription until now + 3:20h, so
we'll kick you out exactly at that point in time".

One could do this using renegotiation/reauth, or by kicking from the
management interface, but if you want to have this integrated with 
radius servers telling you "this user needs to go in 3:20h", having
something like that avoids having to have a management process around...

gert

-- 

ordex:
ok. Isn't `--session-timeout` a  better name than `--session` ?

ordex:
I took the freedom to recreate the patch (codebase was a bit different) and send it to the ml.

The patch includes the change below
> ok. Isn't `--session-timeout` a better name than `--session` ?

and gets rid of the _dowork() thing that is useless imho.

Closing this PR. Thanks!

PR #28, opened 2015-07-19
systemd: client: use KillMode=mixed

blueyed:
The default ("control-group") will prevent the down-root plugin to
communicate with the forked process correctly.

Fixes https://community.openvpn.net/openvpn/ticket/581.
Reported for Debian at: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=792907

### DIFF
--- a/distro/systemd/openvpn-client@.service
+++ b/distro/systemd/openvpn-client@.service
@@ -15,6 +15,7 @@ CapabilityBoundingSet=CAP_IPC_LOCK CAP_NET_ADMIN CAP_NET_BIND_SERVICE CAP_NET_RA
 LimitNPROC=10
 DeviceAllow=/dev/null rw
 DeviceAllow=/dev/net/tun rw
+KillMode=mixed
 
 [Install]
 WantedBy=multi-user.target


blueyed:
Any feedback on this?

It will be in the next Debian upload: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=792907#10.

cron2:
Hi,

On Tue, Sep 15, 2015 at 03:30:50PM -0700, Daniel Hahler wrote:
> Any feedback on this?
> 
> It will be in the next Debian upload: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=792907#10.

Please send patches to the openvpn-devel@lists.sourceforge.net list.  

We don't work with github pull requests (plain and simple: lack of 
developers, and well-defined process "everything that goes in has to
be in the mailing list archive, and is referenced in the commit by 
message ID").

gert

-- 
                                                           //www.muc.de/~gert/

blueyed:
Done.

blueyed:
@cron2 
Thanks for the information / feedback.  I should have remembered this from a previous PR.

It might help to add a CONTRIBUTING file, which would also be displayed by Github (https://github.com/blog/1184-contributing-guidelines).

cron2:
Hi,

On Wed, Sep 16, 2015 at 01:32:41AM -0700, Daniel Hahler wrote:
> It might help to add a CONTRIBUTING file, which would also be displayed by Github (https://github.com/blog/1184-contributing-guidelines).

Thanks for the pointer.  I'll get mattock interested :-) - he's the
project owner on github...

I only break sources and push :-)

gert

-- 
                                                           //www.muc.de/~gert/

blueyed:
The CONTRIBUTING file should apply to the repo itself, and should not be specific to Github.  But it gets displayed when creating issues / PRs on Github, too.
It should probably include a short description how to use `git send-email` to submit your patch etc.

mattock:
I think a CONTRIBUTING file makes sense. I'll put this to my TODO.

cron2:
hi,

On Wed, Sep 16, 2015 at 02:07:42AM -0700, Samuli Seppänen wrote:
> I think a CONTRIBUTING file makes sense. I'll put this to my TODO.

Thanks!

gert
-- 
                                                           //www.muc.de/~gert/

blueyed:
@cron2 
The mailing list post was rejected.  It's a bit cumbersome to subscribe for this!

Can you use the patch from https://patch-diff.githubusercontent.com/raw/OpenVPN/openvpn/pull/28.patch?

mattock:
FYI: CONTRIBUTING.rst file is now in place.

mattock:
A while back we changed our policy regarding GitHub pull requests: we now allow using GitHub for code reviews, even though the final patch still needs to go to the mailing list. Patches from occasional/one-time commiters can be sent to the list by somebody else (e.g. me) to reduce friction.

@dsommers : do you think changing the KillMode is the correct approach? I don't fully understand the [issue](https://community.openvpn.net/openvpn/ticket/581), let alone why the fix fixes it.

dsommers:
Just had a quick glance at this, and it sounds like a plausible solutiuon.  But I have a feeling it hides an issue with the down-root plug-in interface.  If the process down-root should be kicking off when OpenVPN stops gets killed the whole purpose of the plug-in is lost.

I would even say that this script should be kicked off even if OpenVPN is killed/stops running unexpectedly, as this plug-in is often used to clean up system changes the OpenVPN configuration did during the start-up/init phase.

dsommers:
I've finally had some time to look into this now, and read the systemd.kill (5) man page more properly.

This does not look like the proper approach.  I do acknowledge that it will result in all forked processes to be killed - but that is a pure result of them getting the SIGKILL signal.  That should really be avoided.

This is what the man page says about _mixed mode_:

> If set to mixed, the SIGTERM signal is sent to the main process while the subsequent SIGKILL signal is sent to all remaining processes of the unit's control group.

The default mode is _control-group_, which is the proper approach:

> If set to control-group, all remaining processes in the control group of this unit will be killed on unit stop (for services: after the stop command is executed, as configured with ExecStop=)

The default kill signal being sent is SIGTERM.

Which means, if the down-root plug-in leaves an orphaned child process on the system, we need to understand why that happens and fix that.  Changing the _KillMode_ is just a nasty workaround.

blueyed:
Fair enough!
I do not remember the details and what I've meant with "will prevent the down-root plugin to
communicate with the forked process correctly.".

Please be aware that this appears to be in Debian already though - sorry.

blueyed:
Well, from the description at https://community.openvpn.net/openvpn/ticket/581 it becomes clearer again.
Apparently the issue is that all processes of the control group are killed at once.

So if `mixed-mode` sounds too harsh, then `process` should be used probably?1

>            If set to control-group, all remaining processes in the control group of this
           unit will be killed on unit stop (for services: after the stop command is
           executed, as configured with ExecStop=). If set to process, only the main
           process itself is killed. If set to mixed, the SIGTERM signal (see below) is
           sent to the main process while the subsequent SIGKILL signal (see below) is
           sent to all remaining processes of the unit's control group.

dsommers:
That does initially sound somewhat better.  Yesterday evening, right before closing this ticket, I sent a patch to the -devel mailing list which includes proposed changes from PR #22.  Could you please have a look and see if that will work better?  If the issue is still present, then we should test and see if KillMode=process is solving this issue.

PR #30, opened 2015-08-03
net config: allow IP spoofing by clients

StevenVanAcker:


dsommers:
Can you please elaborate more on why we should add this patch?  What is the use case you see?

We are trying to be quite conservative adding more options to OpenVPN, as OpenVPN already got far too many of them already.  But with the proper reasoning we might accept it.  Currently I see this more like a hack than truly useful feature.

neuhalje:
@dsommers This PR seems to be abandoned, can you please close it?

StevenVanAcker:
the reason behind this patch is that I want to have a VPN network where clients are allowed to spoof IP addresses, in order to simulate a real network (it's for a lab).
I can understand if you don't want to have this in the vanilla OpenVPN, although it would be
more convenient for me not having to repatch OpenVPN on every new release

schwabe:
if you just need more than one ip adress per client you can look into iroute

chipitsine:
if you run openvpn in "tap" mode, i.e. L2, can you spoof ?

StevenVanAcker:
that doesn't qualify as spoofing :)

PR #33, opened 2015-08-20
Clear input buffer properly in get_console_input_systemd()

syskill:
The CLEAR macro won't work here because the size of the input buffer is not known. This causes random garbage to be sent as the static challenge response, usually resulting in an error on the Access Server like
```
VPN Auth Failed: "'utf8' codec can't decode byte 0xd0 in position 9: invalid continuation byte: omi/auth:454,auth/domuserpass:97,auth/pwresp:21,util/types:21,encodings/utf_8:16 (exceptions.UnicodeDecodeError)" [None]
```
Strangely, I could not reproduce the error when the response was six characters, as is the case for Google Authenticator codes. The bug only reared its ugly head to us when a user tried to authenticate with an eight-digit backup code.

### DIFF
--- a/src/openvpn/console.c
+++ b/src/openvpn/console.c
@@ -172,7 +172,7 @@ get_console_input_systemd (const char *prompt, const bool echo, char *input, con
   if ((std_out = openvpn_popen (&argv, NULL)) < 0) {
 	  return false;
   }
-  CLEAR (*input);
+  memset (input, 0, capacity);
   if (read (std_out, input, capacity) != 0)
     {
        chomp (input);


dsommers:
You have absolutely found an issue here.  The CLEAR() macro is defined as 
`#define CLEAR(x) memset(&(x), 0, sizeof(x))`.  That results in `sizeof(*input)` returning 1 when I check it via gdb.  So it takes the size of a `char` in an array of chars, which is expected when you pass *input - it dereferences the pointer.

In addition memset() "works" and sets the first byte to 0, because `&(*input)` ends up pointing at the address of `input`.

Your patch solves this particular issue.  But I wonder if we have far more issues because of this.  CLEAR() is used many places, including dereferencing the buffer pointer.  So I think we need to see this in a much bigger scope.  And maybe consider if the CLEAR() macro needs to be improved instead.

syskill:
The CLEAR macro works as long as the size of the object to be cleared is known to the compiler, so it's fine for clearing one object, an array of objects, or a pointer to a single object. It's only in cases like this (a pointer to an array) that it will fail. Unfortunately there is no way to improve the macro to cover this last case. Such is the lot of a C programmer. :disappointed:

I spot-checked some of the other uses of CLEAR, and they look ok.

syskill:
Ping?

schwabe:
To get your patch reviewed and accepted you need to submit it to the openvpn-devel mailing list, see also the CONTRIBUTING.rst for more information.

mattock:
After a recent discussion the development team agreed that using PRs for code review is fine. The final, reviewed patch will still be sent to the list.

@dsommers : any further thoughts on this?

cron2:
Hi,

On Thu, Aug 20, 2015 at 03:49:27PM -0700, Ben Slusky wrote:
> The CLEAR macro won't work here because the size of the input buffer is not known. This causes random garbage to be sent as the static challenge response, usually resulting in an error on the Access Server like

This one can be closed now :-)

commit d09fbf958f1c0b15372b3e87d784ae666b91a96b
Author: Selva Nair <selva.nair@gmail.com>
Date:   Wed Apr 13 23:53:33 2016 -0400

    Ensure input read using systemd-ask-password is null terminated


diff --git a/src/openvpn/console.c b/src/openvpn/console.c
index e1d46c4..86331a1 100644
--- a/src/openvpn/console.c
+++ b/src/openvpn/console.c
@@ -172,8 +172,9 @@ get_console_input_systemd (const char *prompt, const bool echo, char *input, con
   if ((std_out = openvpn_popen (&argv, NULL)) < 0) {
          return false;
   }
-  CLEAR (*input);
-  if (read (std_out, input, capacity) != 0)
+
+  memset (input, 0, capacity);
+  if (read (std_out, input, capacity-1) > 0)
     {
        chomp (input);
        ret = true;

gert

-- 
                                                           //www.muc.de/~gert/

PR #37, opened 2015-11-03
Compiling only entire statements with preprocessor conditionals.

flaviommedeiros:
A suggestion to compile entire statements and expressions, as suggested by code style guidelines from the Linux Kernel and practitioners.

- https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/CodingStyle#n892
- https://www.cqse.eu/en/blog/living-in-the-ifdef-hell/

It might improve code understanding, maintainability and error-proneness.

### DIFF
--- a/src/openvpn/socket.c
+++ b/src/openvpn/socket.c
@@ -802,14 +802,15 @@ create_socket_udp (struct addrinfo* addrinfo, const unsigned int flags)
       else if (addrinfo->ai_family == AF_INET6 )
         {
 #ifndef IPV6_RECVPKTINFO /* Some older Darwin platforms require this */
-          if (setsockopt (sd, IPPROTO_IPV6, IPV6_PKTINFO,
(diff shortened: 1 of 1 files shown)


schwabe:
While I agree on that changing code style might be could idea, this commit only changes a tiny part (one instance) of the ifdefs which are violating this rule. Only a single change might leave us developers more confused than anything else. (the if (failed) style is not is not really used in other parts)

Please see also the CONTRIBUTING.rst file for more information about contributing changes.

cron2:
Thanks for the cleanup attempt.  We discussed this, but decided that we're not going to change these code bits right now - we know that they work, and a change there would need very close scrutiny to ensure no combination of features gets broken (= review time needed).

If such code gets changed due to other changes, we usually do try to cleanup the #ifdefs, preferrably removing the whole compile-time option altogether :-)

PR #38, opened 2015-11-30
netconfig: fixed ipv6 netbits to 126 (4 hosts)

MartinVerges:
I tryed to configure a IPv6 Tunnel Network with 4 IP's and due to the wrong 124 Bit limit, it is not possible at the moment. Therfore I would suggest the fix to set maximum netmask to 126 Bits, the same as in IPv4 a /30 represents.

### DIFF
--- a/src/openvpn/options.c
+++ b/src/openvpn/options.c
@@ -4436,9 +4436,9 @@ add_option (struct options *options,
       if ( get_ipv6_addr( p[1], NULL, &netbits, msglevel ) &&
            ipv6_addr_safe( p[2] ) )
         {
-	  if ( netbits < 64 || netbits > 124 )
+	  if ( netbits < 64 || netbits > 126 )
 	    {
-	      msg( msglevel, "ifconfig-ipv6: /netbits must be between 64 and 124, not '/%d'", netbits );
+	      msg( msglevel, "ifconfig-ipv6: /netbits must be between 64 and 126, not '/%d'", netbits );
 	      goto err;
 	    }
 


mattock:
@cron : any comment on this?

MartinVerges:
would be nice if you could accept it to fix that bug

cron2:
Hi,

On Mon, Nov 30, 2015 at 02:50:01AM -0800, Martin Verges wrote:
> I tryed to configure a IPv6 Tunnel Network with 4 IP's and due to the wrong 124 Bit limit, it is not possible at the moment. Therfore I would suggest the fix to set maximum netmask to 126 Bits, the same as in IPv4 a /30 represents.

NAK.  There are enough bits in IPv6.

gert
-- 
                                                           //www.muc.de/~gert/

cron2:
Hi,

On Thu, Sep 15, 2016 at 03:57:52AM -0700, Martin Verges wrote:
> would be nice if you could accept it to fix that bug

It is not a bug, it's a design choice.

gert
-- 
                                                           //www.muc.de/~gert/

MartinVerges:
can you explain that please? Whats the matter of simply not using the bits 125 and 126? there are only 2 hosts in point-to-point tunnels, so a netmask wich only allows 2 hosts are sufficient from my point of view.

A long time ago someone told the world, there are enough ips and caused the current issues. Yes IPv6 is very big but thats no reason to do it wront or is it? So please explain it, because I would like to understand the "design choice".

cron2:
Hi,

On Thu, Sep 15, 2016 at 01:15:07PM -0700, Martin Verges wrote:
> can you explain that please? Whats the matter of simply not using the bits 125 and 126? there are only 2 hosts in point-to-point tunnels, so a netmask wich only allows 2 hosts are sufficient from my point of view.
> 
> A long time ago someone told the world, there are enough ips and caused the current issues. Yes IPv6 is very big but thats no reason to do it wront or is it? So please explain it, because I would like to understand the "design choice".

IETF says every link (even point to point) should be a /64.

We've loosened this up to permit up to /112, which gives you 281474976710656
subnets in a single /64.

gert

-- 
                                                           //www.muc.de/~gert/

MartinVerges:
RFC 5375 states: "126-bit subnet prefixes are typically used for point-to-point links similar to a the IPv4 address-conservative /30 allocation for point-to-point links.

MartinVerges:
maybe you got some time to read the rfc 5375, and can now explain your design choice over the rfc/ietf. I'm still wondering what's the reason for this.

MartinVerges:
still nothing new?

schwabe:
after you trying to educate us with "Read the rfc you stupid and explain yourself" albeit in much nicer words I lost personally lost interest in fixing this.

cron2:
Hi,

On Fri, Nov 25, 2016 at 03:33:29AM -0800, Martin Verges wrote:
> still nothing new?

Busy with 2.4 release.

I'm going to revisit tunnel / ip handling after 2.4 is done (and all
the "stuff not working as documented" bugs have been fixed).

gert

-- 
                                                           //www.muc.de/~gert/

PR #40, opened 2015-12-07
Fix CR prompt with --auth-user-pass & support reading response.

WayneD:
Fixes a bug where there is no prompt for a dynamic (or static) challenge response when combining "--auth-retry interact" (or --static-challenge) with --auth-user-pass and the password is read from the file. It also extends the reading from the file to allow a response to be provided on the line after the password (which is nice for "push", "phone", etc.).

### DIFF
--- a/doc/openvpn.8
+++ b/doc/openvpn.8
@@ -3800,7 +3800,13 @@ over the client's routing table.
 Authenticate with server using username/password.
 .B up
 is a file containing username/password on 2 lines. If the
-password line is missing, OpenVPN will prompt for one.
+password line is missing or empty, OpenVPN will prompt for one.
+The third line of the file can be used to enter a response to
+a challenge, such as "push", "phone", etc., which can be used
+to eliminate an extra prompt for an unchanging text if your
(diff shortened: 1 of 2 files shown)


cron2:
Hi,

On Mon, Dec 07, 2015 at 02:17:40PM -0800, Wayne Davison wrote:
> Fixes a bug where there is no prompt for a dynamic (or static) challenge response when combining "--auth-retry interact" (or --static-challenge) with --auth-user-pass and the password is read from the file. It also extends the reading from the file to allow a response to be provided on the line after the password (which is nice for "push", "phone", etc.).
> 
>   https://github.com/OpenVPN/openvpn/pull/40

Two things:

 - is this based on git master?  (It would be good if it were, because the
   whole --auth-user-pass thing got changed quite a bit in recent weeks)

 - please submit by git send-mail to openvpn-devel@lists.sourceforge.net
   as we do not process pull requests

thanks,

gert
-- 
                                                           //www.muc.de/~gert/

schwabe:
See also CONTRIBUTING.rst

PR #41, opened 2016-01-10
Fix build with libressl

olesalscheider:


leonklingele:
I built openvpn against libressl, but did not experience any issues.
Can you provide additional information about your setup?

olesalscheider:
Without this patch, I get:

ssl_openssl.c: In function 'tls_ctx_check_cert_time':
ssl_openssl.c:347:10: warning: implicit declaration of function 'SSL_CTX_get0_certificate' [-Wimplicit-function-declaration]
   cert = SSL_CTX_get0_certificate (ctx->ctx);
[...]
ssl_openssl.o: In function `tls_ctx_check_cert_time':
ssl_openssl.c:(.text+0xb00): undefined reference to `SSL_CTX_get0_certificate'

This is on Exherbo with libressl 2.2.5.

leonklingele:
You're right, fails to build for me as well with latest version 2.3.10.
2.3.10 worked fine.

cron2:
Hi,

On Sun, Jan 10, 2016 at 03:00:34AM -0800, Niels Ole Salscheider wrote:
> -- Patch Links --
> 
> https://github.com/OpenVPN/openvpn/pull/41.patch
> https://github.com/OpenVPN/openvpn/pull/41.diff

Besides the fact that we don't do pull requests (please send patches
with git send-email to openvpn-devel@lists.sourceforge.net for review
and merging) - I wonder why LibreSSL is defining a high-enough
OPENSSL_VERSION_NUMBER if it is not providing the full API of
OpenSSL 1.0.2?

I think this should filed as a LibreSSL bug...

gert

-- 
                                                           //www.muc.de/~gert/

olesalscheider:
I have posted the patch to the list.

I guess that LibreSSL do not want to promise to support all of the API of OpenSSL (they even removed some older stuff so they can't offer a superset) or to introduce functions in the same order as OpenSSL.
It is therefore not clear how they should choose the "right" value for OPENSSL_VERSION_NUMBER and they decided to just report the version of LibreSSL.

leonklingele:
Fixed in https://github.com/OpenVPN/openvpn/commit/9dfc2309c6b4143892137844197f5f84755f6580

PR #42, opened 2016-01-30
Fix signed integer shift overflow

mmcco:
OpenVPN uses two loops of this form:

for (i = 1; i; i <<= 1)

where i is an int. This depends on shifting into and out of the sign
bit, both of which are undefined.

This code block has been copied around from original BSD code. I've
already fixed the instances present in OpenBSD:

https://marc.info/?l=openbsd-tech&m=145377854103866&w=2

My analysis there applies here as well.

### DIFF
--- a/src/openvpn/route.c
+++ b/src/openvpn/route.c
@@ -3128,7 +3128,8 @@ get_default_gateway (struct route_gateway_info *rgi)
   struct gc_arena gc = gc_new ();
   struct rtmsg m_rtmsg;
   int sockfd = -1;
-  int seq, l, pid, rtm_addrs, i;
+  int seq, l, pid, rtm_addrs;
+  unsigned int i;
   struct sockaddr so_dst, so_mask;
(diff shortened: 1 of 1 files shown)


cron2:
Hi,

On Sat, Jan 30, 2016 at 09:06:37AM -0800, Michael McConville wrote:
> OpenVPN uses two loops of this form:
> 
> for (i = 1; i; i <<= 1)
> 
> where i is an int. This depends on shifting into and out of the sign
> bit, both of which are undefined.

It seems to be well-behaved enough, but I agree that changing this to
an unsigned bit type is better.

> My analysis there applies here as well.
> 
>   https://github.com/OpenVPN/openvpn/pull/42

As we don't do pull requests, please send a patch to 
openvpn-devel@lists.sourceforge.net (see "CONTRIBUTING.rst").

gert
-- 
                                                           //www.muc.de/~gert/

PR #43, opened 2016-04-27
add unit test support with cmocka

neuhalje:
This set of patches adds unit test support to openvpn. The test framework used is [cmocka](https://cmocka.org/)

* cmocka source is added as submodule in `vendor` because a package depenency is not feasible  (e.g. not available on all platforms)
* cmocka is automatically build during `make check`.
* the patchset contains tests for `searchandreplace` as an example

Tested on MacOS & Ubuntu Trusty

CAVE: 
* cmocka needs cmake to be installed


neuhalje:
I put the PR into a new branch (and did some polishing). The new PR can be found as #44

PR #44, opened 2016-04-29
Add unit tests (replaces PR#43)

neuhalje:
This set of patches adds unit test support to openvpn. The test framework used is [cmocka](http://cmocka.org)

```bash
# after first checkout
git submodule init && git submodule update

# run tests, compile cmocka if needed
make check
```

* cmocka source is added as submodule in vendor because a package depenency is not feasible (e.g. not available on all platforms)
* cmocka is automatically build during make check.
* the patchset contains tests for searchandreplace as an example
* Tested on MacOS & Ubuntu Trusty

CAVE:

* cmocka needs cmake to be installed

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1198,6 +1198,19 @@ sampledir="\$(docdir)/sample"
 AC_SUBST([plugindir])
 AC_SUBST([sampledir])
 
+VENDOR_SRC_ROOT="\$(abs_top_srcdir)/vendor/"
+VENDOR_DIST_ROOT="\$(abs_top_builddir)/vendor/dist"
+VENDOR_BUILD_ROOT="\$(abs_top_builddir)/vendor/.build"
+AC_SUBST([VENDOR_SRC_ROOT])
+AC_SUBST([VENDOR_BUILD_ROOT])
+AC_SUBST([VENDOR_DIST_ROOT])
+
+TEST_LDFLAGS="-lcmocka -L\$(abs_top_builddir)/vendor/dist/lib -Wl,-rpath,\$(abs_top_builddir)/vendor/dist/lib"
+TEST_CFLAGS="-I\$(top_srcdir)/include -I\$(abs_top_builddir)/vendor/dist/include"
+
+AC_SUBST([TEST_LDFLAGS])
+AC_SUBST([TEST_CFLAGS])
+
 AC_CONFIG_FILES([
 	version.sh
 	Makefile
@@ -1216,6 +1229,9 @@ AC_CONFIG_FILES([
 	src/plugins/auth-pam/Makefile
 	src/plugins/down-root/Makefile
 	tests/Makefile
+        tests/unit_tests/Makefile
+        tests/unit_tests/example_test/Makefile
+        vendor/Makefile
 	sample/Makefile
 	doc/Makefile
 ])
--- a/configure.ac
+++ b/configure.ac
@@ -1230,6 +1230,8 @@ AC_CONFIG_FILES([
 	src/plugins/down-root/Makefile
 	tests/Makefile
         tests/unit_tests/Makefile
+        tests/unit_tests/plugins/Makefile
+        tests/unit_tests/plugins/auth-pam/Makefile
         tests/unit_tests/example_test/Makefile
         vendor/Makefile
 	sample/Makefile
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -12,6 +12,8 @@
 MAINTAINERCLEANFILES = \
 	$(srcdir)/Makefile.in
 
+SUBDIRS = unit_tests
+
 test_scripts = t_client.sh t_lpback.sh t_cltsrv.sh
 
 TESTS_ENVIRONMENT = top_srcdir="$(top_srcdir)"
@@ -20,4 +22,3 @@ TESTS = $(test_scripts)
 dist_noinst_SCRIPTS = \
 	$(test_scripts) \
 	t_cltsrv-down.sh
-
--- a/tests/unit_tests/.gitignore
+++ b/tests/unit_tests/.gitignore
@@ -1,0 +1,1 @@
+*_testdriver
--- a/tests/unit_tests/README.md
+++ b/tests/unit_tests/README.md
@@ -1,0 +1,40 @@
+Unit Tests
+===========
+
+This directory contains unit tests for openvpn. New features/bugfixes should be written in a test friendly way and come with corresponding tests.
+
+Run tests
+----------
+
+Tests are run by `make check`. A failed tests stops test execution. To run all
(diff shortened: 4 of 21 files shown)


chipitsine:
openvpn somehow does not close PR on github, so people usually open PR ... wait for some time ... someone respondes "we do not use PR here, please do so and so"

have a look at other PR

it is really strange, I've no idea why PR are not closed for this project

neuhalje:
@chipitsine This is really strange.  But thanks for the pointer: I  know that patches are managed via the list, I already send the patches.

mattock:
Some time back we decided to allow use PRs for code review. The reviewed patch / set of patches will be sent to openvpn-devel mailing list and merged from there. Sending the patch can be handled by a core team member (e.g. me), if the author does not want to do that.

mattock:
Regarding the patch... I noticed that the Git submodule is inside a subdirectory. If I recall correctly, older Git versions only allow submodules at the root of the Git repository. I wonder if this is going to cause issues in practice?

I'll test this patch and see how it works out on Fedora 23 and what caveats there are in practice, then get back to this.

neuhalje:
@mattock git submodule with folder works on Ubuntu Precise with git version 1.7.9.5  (released [march 2012](https://github.com/git/git/releases/tag/v1.7.9.5))

neuhalje:
FYI: PR #44 can be used to test unit testing for different platforms; just make sure to install cmake (not included in #44)

syzzer:
Thanks.  This look promising, but I noticed that out-of-source builds fail.

I.e. the following should work, but currently doesn't:

    (git clone --recursive https://github.com/neuhalje/openvpn.git --branch add_unit_tests && cd openvpn && autoreconf -vi)
    mkdir openvpn-test-build
    cd openvpn-test-build
    ../openvpn/configure
    make check

neuhalje:
@syzzer  Good catch! Pushed a fix, mind to re-test for double check?

tests/unit_tests/example_test/Makefile.am
```diff
@@ -0,0 +1,13 @@
+AUTOMAKE_OPTIONS = foreign
+
+check_PROGRAMS = testdriver testdriver2
```
syzzer: More descriptive names would be good.  `make check` now reports 'PASS: testdriver', which is not very informative.

tests/unit_tests/example_test/Makefile.am
```diff
@@ -0,0 +1,13 @@
...
+
+TESTS = testdriver testdriver2
```
syzzer: Why not `TEST = $(check_PROGRAMS)`?  I'd prefer to only have to maintain a single list :)

tests/Makefile.am
```diff
@@ -21,3 +23,7 @@ dist_noinst_SCRIPTS = \
 	$(test_scripts) \
 	t_cltsrv-down.sh
 
+.PHONY:	unit_tests
+unit_tests:
```
syzzer: This seems to not be needed.  `SUBDIRS = unit_tests` suffices.  (Also, there is no 'unit_tests' target in unit_tests/Makefile.am.)

tests/unit_tests/plugins/auth-pam/Makefile.am
```diff
@@ -0,0 +1,11 @@
+AUTOMAKE_OPTIONS = foreign
+
+check_PROGRAMS = testdriver
```
syzzer: This should probably be wrapped in `if ENABLE_PLUGIN_AUTH_PAM` (like in src/plugins/auth-pam/Makefile.am), so that the tests only run in we are actually compiling the plugin.  (Why not always run plugin tests? Well, perhaps libpam is not available, but we do want to run `make check`.)

tests/unit_tests/example_test/Makefile.am
neuhalje: Maybe rename them to _example_testdriver_ and _example_testdriver2_ because they are really just that. We can even delete them when we have enough "real" tests that make a template superfulous.

tests/unit_tests/example_test/Makefile.am
neuhalje: Good idea!

tests/unit_tests/README.md
```diff
@@ -0,0 +1,27 @@
+Unit Tests
+==========
+
+This directory contains unit tests for openvpn.
+
+Run tests
+------------
```
syzzer: Many of these lines' lengths do not match the title lengths.

syzzer:
Yes, works as expected now.

I'm going slightly pedantic with the review, but since this will be acting as an example that everyone will be copy-pasting around, it's extra important to get it fully right.

tests/unit_tests/example_test/Makefile.am
syzzer: Yes, or testdriver_example1, so that you can still use your catch-all in .gitignore.  I was more talking in general terms, as the auth-pam test was also called 'testdriver'.

tests/Makefile.am
neuhalje: I'll remove it (must have slipped by)

neuhalje:
@syzzer  Bad examples are the worst, so I totally agree with a nitty gritty review :-)

neuhalje:
Just give me a heads up when you are through, and I'll send in the patches to the list

neuhalje:
Testing on Solaris 11.3:
* Testdrivers must not contain underscore in name
* Besides that: works

Update follows

configure.ac
```diff
@@ -1216,6 +1229,11 @@ AC_CONFIG_FILES([
 	src/plugins/auth-pam/Makefile
 	src/plugins/down-root/Makefile
 	tests/Makefile
+        tests/unit_tests/Makefile
+        tests/unit_tests/plugins/Makefile
+        tests/unit_tests/example_test//Makefile
```
syzzer: Is this double slash on purpose?

tests/unit_tests/.gitignore
```diff
@@ -0,0 +1,4 @@
+test_driver*
+testdriver*
+example_test_driver*
+example_testdriver*
```
syzzer: This .gitignore should either move into example_test, or be made generic.  I'd suggest the second. E.g. ignore *_testdriver and use the convention that all testdriver programs are called that way (which would mean renaming example_testdriver2 to example2_testdriver). That way not every subdir will need it's own .gitignore.

syzzer:
Two final nits, but after handling those this should be ready for the list :)  Not sure if you were planning to, but I think it's best to sqaush the fixes back into the two original commits.

Thanks for bearing with me.

cron2:
has this been merged fully, so we can close the PR?

syzzer:
Yes, merged!

PR #45, opened 2016-04-29
testing: Add vagrant based integration tests

neuhalje:
This patch will add test servers to use with t_client.sh. This will make testing on different systems much easier. Bonus: testing on a laptop is a breeze by "one command integration tests".

_Example_

``` bash
./run_integration.sh ubuntu_precise64
# does the following
# - start an Ubuntu Precise VM
# - Copy your current source code to the VM
# - configure & make on the VM
# - Run the server on the VM
# - Run t_client.sh *with the REMOTE_HOST pointing to the VM*
# - The outputs of client & server are colorized AND interleaved for better diagnosis
# - Tear down the server
```

Virtual machines are defined by [Vagrant](https://www.vagrantup.com) scripts. Virtualbox is used for virtualisation.

Each of the virtual machines is configured to compile openvpn and run a server.

Inside the vm a openvpn server can be launched to run `t_client.sh` tests against it. Sample launcher scripts are provided.

screen shot 2016-05-09 at 16 17 18

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -309,11 +309,17 @@ AC_ARG_WITH(
 )
 
 
+# Test if ld supports '--wrap' whcih currently is only supported on Linux
+# http://www.mockator.com/projects/mockator/wiki/Wrap_Functions
+AM_CONDITIONAL([SUPPORT_LD_WRAP], [ false ])
+
 AC_DEFINE_UNQUOTED([TARGET_ALIAS], ["${host}"], [A string representing our host])
 case "$host" in
 	*-*-linux*)
 		AC_DEFINE([TARGET_LINUX], [1], [Are we running on Linux?])
 		AC_DEFINE_UNQUOTED([TARGET_PREFIX], ["L"], [Target prefix])
+                # It might be better to actually run ld with --wrap
+                AM_CONDITIONAL([SUPPORT_LD_WRAP], [ true ])
 		;;
 	*-*-solaris*)
 		AC_DEFINE([TARGET_SOLARIS], [1], [Are we running on Solaris?])
@@ -1262,6 +1268,10 @@ TEST_CFLAGS="-I\$(top_srcdir)/include -I\$(abs_top_builddir)/vendor/dist/include
 
 AC_SUBST([TEST_LDFLAGS])
 AC_SUBST([TEST_CFLAGS])
+
+# Test if ld supports '--wrap' whcih currently is only supported on Linux
+# http://www.mockator.com/projects/mockator/wiki/Wrap_Functions
+AM_CONDITIONAL([SUPPORT_LD_WRAP], [TARGET_LINUX])
 
 # Check if cmake is available and cmocka git submodule is initialized,
 # needed for unit testing
--- a/tests/server/.gitignore
+++ b/tests/server/.gitignore
@@ -1,0 +1,1 @@
+.vagrant/
--- a/tests/server/README.md
+++ b/tests/server/README.md
@@ -1,0 +1,45 @@
+VM definitions for test servers
+================================
+
+Virtual machines defined by [Vagrant](https://www.vagrantup.com) scripts. Virtualbox is used for virtualisation.
+
+Each of the virtual machines is configured to compile openvpn and run a server.
+
+Inside the vm a openvpn server can be launched to run `t_client.sh` tests against it.
+
+Use Cases:
+---------------
+ 
+### Start the VM:
+  * `vagrant up` from inside the VMs folder (e.g. `centos7`)
+
+### Login to  the VM:
+  * `vagrant ssh`
+
+### Synchronize local changes into the VM & rebuild the server
+  * (insinde the vm)
+  *  `/scripts/sync_codebase.sh`
+  *  `/scripts/compile.sh`
+
+### Start the server for  `t_client` tests
+  * (outside of the vm) Modify `t_client.rc` ( `REMOTE_IP`, `REMOTE_PORT`, `SERVER_KEY`, `SERVER_CERT`
+  * (inside the vm) run `/scripts/sync_codebase.sh` (to sync the changed `t_client.rc`)
+  * (inside the vm) run `/scripts/compile.sh`
+  * (inside the vm) run `/scripts/start_server.sh`
+
+VM Layout
+-----------
+ * The original source tree is mounted at /openvpn
+ * The source tree used for compilation is copied to `~vagrant/openvpn`
+ * Scripts are installed in the /scripts directory
+   * `sync_codebase.sh` copies the codebase from the host into ~/openvpn ( so that it can be compiled for the VMs architecture/OS)
+   * `rebuild.sh` builds the server inside the VM (at ~/openvpn)
+   * `start_server.sh` starts the server with a configuration from t_client.rc
+
+provided VMs
+--------------
+
+| Name                     | external IPv4   | external IPv6              |
+|--------------------------|-----------------|----------------------------|
+| ubuntu_trusty64          | 192.168.33.66   | IPv4 only                  |
+| ubuntu_precise64         | 192.168.33.67   | IPv4 only                  |
--- a/tests/server/README.md
+++ b/tests/server/README.md
@@ -7,28 +7,29 @@ Each of the virtual machines is configured to compile openvpn and run a server.
 
 Inside the vm a openvpn server can be launched to run `t_client.sh` tests against it.
 
-Use Cases:
----------------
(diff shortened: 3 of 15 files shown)


mattock:
I've used Vagrant quite a bit in the past, so I'll take care of reviewing this. Right now I don't have VirtualBox installed, so the review will take a bit of time.

mattock:
I tested this a bit with latest Vagrant+Virtualbox and most of it seemed to work fine out of the box - excellent work!

I'm just wondering what kind of test setup we want exactly... right now, the "real" test server runs several OpenVPN server instances configured differently and listening on different ports. Each of the clients (buildslaves) connect to each of the server instances in turn. This is something we might (or might not) want to replicate here.

Another option would be to focus on building and running the server-side code on different VM instances, each running a different OS, and run the tests to ensure that the _virtual host_ can connect to each of them. There could be several differently configured OpenVPN server instances on each of the VMs. Is this what you're trying achieve here?

In any case, I think we should strive for maximum automation, so that any developer is able to recreate the t_client.sh test environment with minimal tinkering and knowledge about t_client.sh. Which brings me to a couple of issues I noticed:

* It seems that the line _RUN_SUDO=sudo_ in t_client.rc needs to be enabled: at least on Fedora doing a "sudo -s" changes $HOME to /root, which means that $HOME/.vagrant can no longer be found.
* The _EXCEPT_*_ values in t_client.rc-sample are not correct for the Vagrant VMs, which causes t_client.sh fail, even though OpenVPN connection is established correctly

What about having a tailored _t_client.rc-vagrant_ file that fixes the two issues above? Making the VPN IP addresses predictable probably takes some effort, but would be worth it imho.

This is probably unrelated to this patch, but I noticed that the "stopping OpenVPN" part in t_client.sh hangs (indefinitely?) on Fedora 23.

neuhalje:
Hi

> I tested this a bit with latest Vagrant+Virtualbox and most of it seemed to work fine out of the box - excellent work!
> 
Thanks! And thank you for testing!
> I'm just wondering what kind of test setup we want exactly... right now, the "real" test server runs several OpenVPN server instances configured differently and listening on different ports. Each of the clients (buildslaves) connect to each of the server instances in turn. This is something we might (or might not) want to replicate here.
> 
basically “n Client Systems x n Server Systems”? This is what I see as a good case for a “is this ready for a patch submit” test.
> Another option would be to focus on building and running the server-side code on different VM instances, each running a different OS, and run the tests to ensure that the virtual host can connect to each of them. There could be several differently configured OpenVPN server instances on each of the VMs. Is this what you're trying achieve here?
> 
basically “1 Client System x n Server Systems”?

Both are viable routes that might not be exclusive (e.g. maybe they are more or less the same with a slightly different configuration)

IMHO the use cases are:

1) Iterative development (“1 x 1”): One VM as server or client endpoint. The other side on the development machine.

2) Feature test (“1 x n”): Multiple VMs as server or client endpoint. The other side on the development machine.

3) Final test (“n x n”):  Multiple VMs as server AND client endpoint. 

Use case #1 is a MUST. Use case #2 is a SHOULD, and #3 its a COULD

My high level goals:

1) Make it easy for developers to set up different test & build machines to test new features against. (hypothesis: done with this PR, except for maybe some bugs)

2) Make integration with existing test tooling as easy as possible (still not quite there with this PR).

3) Make it easy to separate”Test Definition” from “Execution Environment” (still not quite started with this PR).
    I.e. a test should consist of a client configuration, a server configuration, and (a maybe fixed) test function. 
    This is basically what t_client.sh does right now with the addition of automatic address configuration, and automatic server launches.

> In any case, I think we should strive for maximum automation, so that any developer is able to recreate the t_client.sh test environment with minimal tinkering and knowledge about t_client.sh. Which brings me to a couple of issues I noticed:
> 
Yes!

> It seems that the line RUN_SUDO=sudo in t_client.rc needs to be enabled: at least on Fedora doing a "sudo -s" changes $HOME to /root, which means that $HOME/.vagrant can no longer be found.
—> Bug
> The EXCEPT*_ values in t_client.rc-sample are not correct for the Vagrant VMs, which causes t_client.sh fail, even though OpenVPN connection is established correctly
—> Bug

> What about having a tailored t_client.rc-vagrant file that fixes the two issues above? Making the VPN IP addresses predictable probably takes some effort, but would be worth it imho
> 
Sounds good. I’ll try to improve the scripts next weekend!
> This is probably unrelated to this patch, but I noticed that the "stopping OpenVPN" part in t_client.sh hangs (indefinitely?) on Fedora 23.
> 
> 
—> Bug

Next iteration :-)

Cheers
Jens

mattock:
> basically “n Client Systems x n Server Systems”? This is what I see as a good case for a “is this ready for a patch submit” test.

Right now we do this after the patch is in Git "master", because buildbot polls the "master" branch. We could and intend to add more testing branches that allow us to trigger builds _before_ pushing anything to master.

> IMHO the use cases are:
> 
> 1) Iterative development (“1 x 1”): One VM as server or client endpoint. The other side on the development machine.

This is a good start and a reasonable sanity test. We can implement 2) and 3) as needed.

> Use case #1 is a MUST. Use case #2 is a SHOULD, and #3 its a COULD

I agree, as we're probably targeting individual developers with Vagrant. The project as a whole can afford more complex testing systems, as those are built only once and the maintenance cost is thus low.

mattock:
@neuhalje : any progress on fixing the issues that I identified?

One further issue I noticed is the use of #!/bin/bash in some of the scripts. The scripts might include some bashisms, which need to be removed also. While most developers probably run Linux, Vagrant and Virtualbox also run on FreeBSD, which does not have Bash installed by default. So it would be preferable If the scripts were converted to POSIX shell format to avoid complaints later on.

As the last step the commits could probably be squashed to a few larger ones with no ill effects.

mattock:
I'll look into the bashisms to get this moving forward. This PR might benefit from the new t_client.rc EXPECT_IFCONFIG* auto-detection and caching [patch](https://github.com/OpenVPN/openvpn/commits/master) that just got merged.

neuhalje:
Ok, back again. I'm going to try to get the scripts ksh compatible.

neuhalje:
@mattock : Samuli, I have been absent from OpenVPN development for a few, aehm, days. Mayber you could help me out:

I already removed the `EXPECT_IFCONFIG*` statements from `t_client.rc-sample`. Is that enough to implement df0b00c25?

mattock:
@neuhalje : it's been a while since I implemented the IP caching, but two things need to be done:
* Remove EXCEPT_IFCONFIG entries from t_client.rc
* Include t_client_ips.rc in t_client.rc (see [t_client.rc-sample](https://github.com/OpenVPN/openvpn/commit/df0b00c253e41cce9567be79dbd3faa14c60473b#diff-0ada82779e4fb82b2542dddf659b55e9R15)): otherwise t_client.sh will add duplicate EXCEPT_IFCONFIG lines to t_client_ips.rc on every run.

cron2:
Independent from the rest of this, the "run tests needing 'ld --wrap' only where supported" sounds like something we want to have right away...  can you send that one as standalone patch relative to master to the -devel list?  thanks

neuhalje:
I pushed the 'ld --wrap' fix as PR #79 
It still needs some test though

neuhalje:
Gert, any clue on how to run a `ld --wrap`-works test? My autotools are sub-zero

cron2:
Hi,

On Tue, Jan 17, 2017 at 03:13:37PM -0800, Jens Neuhalfen wrote:
> Gert, any clue on how to run a `ld --wrap`-works test? My autotools are sub-zero

My autotools knowledge is at 0.03 or so, so "more than zero because I
had to", but not really.  Selva is the wonderman, though :-)

But we'll get this done together!

gert

-- 
                                                           //www.muc.de/~gert/

chipitsine:
```
[ilia@localhost ~]$ ld --help | grep wrap
  --wrap SYMBOL               Use wrapper functions for SYMBOL
[ilia@localhost ~]$ 
```

neuhalje:
on macos:
```
 ld --help
ld64: For information on command line options please use 'man ld'.
```

chipitsine:
try 'ld --help 2>/dev/null'

neuhalje:
Let's continue the '--wrap' topic on PR#79

chipitsine:
sorry, I did not mention vagrant in the topic

neuhalje:
probably dead by now .. cleaning up

mattock:
This PR was a good idea and I intended to get this merged. But I had never time to do it, nor does it look likely that I ever will  :disappointed:.

PR #47, opened 2016-05-18
fix cppcheck findings

chipitsine:
(both master and release/2.3)

[src/openvpn/error.c:736]: (error) Possible null pointer dereference: null
[src/openvpn/error.c:736]: (error) Null pointer dereference

function "crash" is not used anywhere. so, it's better to remove it.

(only master)

[src/openvpnserv/interactive.c:601]: (error) Memory pointed to by 'addr_row' is freed twice.
[src/openvpnserv/interactive.c:700]: (error) Memory pointed to by 'fwd_row' is freed twice.
[src/openvpnserv/interactive.c:1329]: (error) Common realloc mistake: 'handles' nulled but not freed upon failure

RemoveListItem free addr_row and fwd_row, so there's no need to free them once again after "goto". It is dirty fix, I'd say interactive.c is overcomplicated here (lots of "if"s and "goto"s), so it is better to rewrite it in some straightforward way.

as for realloc, it is common not to handle failures, it is not very dangerous, it is rather some sugar to stop cppcheck from noising.

### DIFF
--- a/src/openvpnserv/interactive.c
+++ b/src/openvpnserv/interactive.c
@@ -597,6 +597,7 @@ HandleAddressMessage (address_message_t *msg, undo_lists_t *lists)
         goto out;
 
       free (RemoveListItem (&(*lists)[address], CmpAddress, addr_row));
+      return err;
     }
 
 out:
@@ -696,6 +697,7 @@ HandleRouteMessage (route_message_t *msg, undo_lists_t *lists)
         goto out;
 
       free (RemoveListItem (&(*lists)[route], CmpRoute, fwd_row));
+      return err;
     }
 
 out:
@@ -1474,9 +1476,13 @@ UpdateWaitHandles (LPHANDLE *handles_ptr, LPDWORD count,
       if (pos == size)
         {
           size += 10;
-          handles = realloc (handles, size * sizeof (HANDLE));
-          if (handles == NULL)
-            return ERROR_OUTOFMEMORY;
+	  void * tmp = realloc (handles, size * sizeof (HANDLE));
+ 	  if (tmp == NULL)
+	  {
+		free(handles);
+		return ERROR_OUTOFMEMORY;
+	  }
+	  handles = tmp;
         }
       handles[pos++] = threads->data;
       threads = threads->next;
--- a/src/openvpn/error.c
+++ b/src/openvpn/error.c
@@ -728,15 +728,6 @@ msg_flags_string (const unsigned int flags, struct gc_arena *gc)
   return BSTR (&out);
 }
 
-#ifdef ENABLE_DEBUG
-void
-crash (void)
-{
-  char *null = NULL;
-  *null = 0;
-}
-#endif
-
 #ifdef WIN32
 
 const char *


cron2:
Hi,

On Wed, May 18, 2016 at 12:32:42AM -0700, Ilya Shipitsin wrote:
> (both master and release/2.3)
> 
> [src/openvpn/error.c:736]: (error) Possible null pointer dereference: null
> [src/openvpn/error.c:736]: (error) Null pointer dereference
> 
> function "crash" is not used anywhere. so, it's better to remove it.

The whole point of crash() is to crash, and it's only there if ENABLE_DEBUG
is defined - so that null pointer dereference is fully intentional.

Currently it indeed has no caller.

> (only master)
> 
> [src/openvpnserv/interactive.c:601]: (error) Memory pointed to by 'addr_row' is freed twice.

I'm not sure I agree with that assessment.  RemoveListItem() returns
item->data, which is not pointer-compared to "addr_row", but memory-compared
("it points to something that contains the same information as addr_row")
by CmpAddress:

static BOOL
CmpAddress (LPVOID item, LPVOID address)
{
  return memcmp (item, address, sizeof (MIB_UNICASTIPADDRESS_ROW)) == 0 ? TRUE :
 FALSE;
}

so having addr_row point at memory location x and item->data point at 
memory location y, but both having the same content(!) would match here
- and then "item->data" would be free()ed (->y), not "addr_row".


> [src/openvpnserv/interactive.c:700]: (error) Memory pointed to by 'fwd_row' is freed twice.

Same thing.

I agree that this code is not really easy to read - so rewriting it
in a way that makes it easier to parse for humans and static analyzers
might be the way forward.  But that doesn't make the existing code
incorrect.

> RemoveListItem free addr_row and fwd_row, 

It does not.  It frees the memory region data->item, which has the same
*content* as the memory region pointed to by addr_row and fwd_row, but
is not the same pointer.

(Technically, it frees only "data", which is the list node that happens
to have data->item pointing to the memory location identical to addr_row /
fwd_row, and the caller free()'s data->item - but in any case, nobody 
free()s what is "ctx" in RemoveListItem().


> [src/openvpnserv/interactive.c:1329]: (error) Common realloc mistake: 'handles' nulled but not freed upon failure

You're working on an older tree - git master line 1329 has no realloc()
anywhere near.  I do think I know which realloc() you are talking about
- Selva has sent a patch for that (now in line 1477) a few days ago, but
withdrawn.  I think he's still looking at this.

gert
-- 
                                                           //www.muc.de/~gert/

src/openvpnserv/interactive.c
```diff
@@ -597,6 +597,7 @@ HandleAddressMessage (address_message_t *msg, undo_lists_t *lists)
         goto out;
 
       free (RemoveListItem (&(*lists)[address], CmpAddress, addr_row));
+      return err;
```
d12fk: This introduces a memory leak, since addr_row is not free()ed.

src/openvpnserv/interactive.c
```diff
@@ -696,6 +697,7 @@ HandleRouteMessage (route_message_t *msg, undo_lists_t *lists)
         goto out;
 
       free (RemoveListItem (&(*lists)[route], CmpRoute, fwd_row));
+      return err;
```
d12fk: This introduces a memory leak, since fwd_row is not free()ed.

src/openvpnserv/interactive.c
```diff
@@ -1474,9 +1476,13 @@ UpdateWaitHandles (LPHANDLE *handles_ptr, LPDWORD count,
       if (pos == size)
         {
           size += 10;
-          handles = realloc (handles, size * sizeof (HANDLE));
-          if (handles == NULL)
-            return ERROR_OUTOFMEMORY;
+	  void * tmp = realloc (handles, size * sizeof (HANDLE));
```
d12fk: This is using tabs instead of spaces.

src/openvpnserv/interactive.c
```diff
@@ -1474,9 +1476,13 @@ UpdateWaitHandles (LPHANDLE *handles_ptr, LPDWORD count,
...
+ 	  if (tmp == NULL)
+	  {
+		free(handles);
```
selvanair: This will lead to a double free when the caller frees handles_ptr which points to the previous value of handles.

selvanair:
@chipitsine
> [src/openvpnserv/interactive.c:601]: (error) Memory pointed to by 'addr_row' is freed twice.
> [src/openvpnserv/interactive.c:700]: (error) Memory pointed to by 'fwd_row' is freed twice.

I too don't see any double free here. cppcheck is likely barking up a wrong tree.
 
> > [src/openvpnserv/interactive.c:1329]: (error) Common realloc mistake: 'handles' nulled but not freed upon failure

@cron2 
> You're working on an older tree - git master line 1329 has no realloc()
> anywhere near.  I do think I know which realloc() you are talking about
> - Selva has sent a patch for that (now in line 1477) a few days ago, but
> withdrawn.  I think he's still looking at this.

The patch was missing a chunk. I did not resend it as it was somewhat cosmetic: viz., "use realloc the right way" even though the original code is fine for some convoluted reasons. Note that memory is allocated in chunks of 10 handles but the required size increases only by 1 per call  (or by 2 for the first call) . So realloc will be called only once at most in the loop, and if it fails, ServiceStartInteractive() has a copy of the original handles pointer. Still, I may resubmit the patch after another look.

cron2:
Hi,

On Fri, May 20, 2016 at 10:24:17AM -0700, Selva Nair wrote:
> @cron2 
> > You're working on an older tree - git master line 1329 has no realloc()
> > anywhere near.  I do think I know which realloc() you are talking about
> > - Selva has sent a patch for that (now in line 1477) a few days ago, but
> > withdrawn.  I think he's still looking at this.
> 
> The patch was missing a chunk. I did not resend it as it was somewhat cosmetic: viz., "use realloc the right way" even though the original code is fine for some convoluted reasons. Note that memory is allocated in chunks of 10 handles but the required size increases only by 1 per call  (or by 2 for the first call) . So realloc will be called only once at most in the loop, and if it fails, ServiceStartInteractive() has a copy of the original handles pointer. Still, I may resubmit the patch after another look.

Thanks for the clarification.  I did look at the code for a few minutes
but couldn't really figure out what all possible error exits are, and 
what *should* happen in that case.

If the current code is "complicated but correct", we can close this PR,
and if you find time, I'll happily stare at a new patch :-)

gert
-- 
                                                           //www.muc.de/~gert/

PR #48, opened 2016-05-18
 fix cppcheck findings

chipitsine:
[src/openvpn/error.c:736]: (error) Possible null pointer dereference: null
[src/openvpn/error.c:736]: (error) Null pointer dereference

function "crash" is not used anywhere. so, it's better to remove it.


cron2:
crash() is already covered in pull request 47, and as commented there, the null pointer derefernce is fully intentional - it does what the name says: crash.

PR #51, opened 2016-05-30
Change unix build instruction

ITJesse:


mattock:
Hi,

The original build instruction are in fact correct when building from release tarballs. When building from Git you _do_ need autoreconf, but you can use this slightly simplified process:

    autoreconf -vi
    ./configure
    make
    make install

We can definitely take the patch if the above changes/concerns are addressed.

selvanair:
To add to @mattock comment, it may be more useful to add a line on how to download and unpack the source tarball. Then ./configure && make && make install is all that is needed. Requiring developer tools such as autoconf for a "quick build" does not look right..

mattock:
I suggest this minor modification:

```
To get the latest release of OpenVPN, go to:

    http://openvpn.net/

To Build and Install,

    tar -zxf openvpn-<version>.tar.gz
    ./configure
    make
    make install

or see the file INSTALL for more info.
```
This makes it clear the instructions are for the tarball, not Git sources.

cron2:
Hi,

On Mon, May 30, 2016 at 12:35:34PM -0700, Samuli Seppänen wrote:
> I suggest this minor modification:

Sounds good.

gert
-- 
                                                           //www.muc.de/~gert/

mattock:
Patch sent to the list.

selvanair:
cd openvpn-<version>
after tar -zxf ...
is missing..

mattock:
Indeed, and this is why using GitHub for review is so much easier than the mailing list :). V3 of this seemingly trivial patch is now on the list.

cron2:
seems as if this has been merged, right?

PR #52, opened 2016-06-04
testing: travis-ci support (for smoke testing)

chipitsine:


.travis.yml
```diff
@@ -0,0 +1,82 @@
...
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install lzo; fi
+
+install:
+  - mkdir ~/bin && ln -s $(which ccache) ~/bin/$CC && export PATH=$HOME/bin:$PATH
+  - if [ ! -d download-cache ]; then mkdir download-cache; fi
+  - if [ ! -f download-cache/mbedtls-${MBEDTLS_VERSION}-apache.tgz ]; then wget -O download-cache/mbedtls-${MBEDTLS_VERSION}-apache.tgz https://tls.mbed.org/download/mbedtls-${MBEDTLS_VERSION}-apache.tgz; fi
+  - tar zxf download-cache/mbedtls-${MBEDTLS_VERSION}-apache.tgz
+  - cd mbedtls-${MBEDTLS_VERSION}
+  - make > build.log 2>&1 || (cat build.log && exit 1)
+  - sudo make install
```
syzzer: I think we should try to avoid using sudo - it excludes some travis options.

Instead, we could set the MBEDTLS_CFLAGS and MBEDTLS_LIBS env variables to point ./configure to our mbedtls build.

.travis.yml
```diff
@@ -0,0 +1,82 @@
...
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update     ; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install lzo; fi
+
+install:
+  - mkdir ~/bin && ln -s $(which ccache) ~/bin/$CC && export PATH=$HOME/bin:$PATH
+  - if [ ! -d download-cache ]; then mkdir download-cache; fi
+  - if [ ! -f download-cache/mbedtls-${MBEDTLS_VERSION}-apache.tgz ]; then wget -O download-cache/mbedtls-${MBEDTLS_VERSION}-apache.tgz https://tls.mbed.org/download/mbedtls-${MBEDTLS_VERSION}-apache.tgz; fi
+  - tar zxf download-cache/mbedtls-${MBEDTLS_VERSION}-apache.tgz
+  - cd mbedtls-${MBEDTLS_VERSION}
+  - make > build.log 2>&1 || (cat build.log && exit 1)
```
syzzer: I understand that the mbedtls build logs are not the most interesting of the travis build, but I'd rather have them available anyway.  That might help in debugging issues.  (And iirc, travis hides the 'install' bit behind a drop-down thingy anyway.)

.travis.yml
chipitsine: I decided to compile mbedtls because there's no ubuntu 14.04 repo :)
otherwise, I would be happy just to install from repo.

ok, I'll rework it without sudo

.travis.yml
syzzer: Can we avoid building mbedtls for openssl builds?  That's nice towards travis-ci (less cpu cycles / disk space), and should speed up openssl builds.

.travis.yml
syzzer: Yes, I understand. Building our own mbedtls makes sense. Looking a bit further down the path (future patch / pull request) I'd even suggest to expand this to run builds against all supported mbedtls / openssl versions.

.travis.yml
```diff
@@ -0,0 +1,83 @@
...
+env:
+  global:
+    - MBEDTLS_VERSION=2.2.1
+    - MBEDTLS_PREFIX=~/opt/mbed
+    - MBEDTLS_LIBS="-L$MBEDTLS_PREFIX/lib -lmbedtls -lmbedx509 -lmbedcrypto"
+    - MBEDTLS_CFLAGS="-I$MBEDTLS_PREFIX/include"
+
+matrix:
+  include:
+    - env: CONFIG="--disable-lzo"
```
syzzer: Why --disable-lzo?  I see you've installed liblzo2-dev, that should be sufficient, right?

syzzer:
Just to be clear - I'm happy you're putting effort into this.  The remarks are just to make sure we get this right before merging.  I for one would really like decent travis support for OpenVPN.

.travis.yml
chipitsine: I picked random line from https://github.com/mattock/openvpn-buildbot/blob/master/master.cfg.sample#L79

the build splits into several configurations, some include lzo, some do not include.

chipitsine:
I can add something like "if $CONFIG does contain 'openssl' then build openssl", but I'm afraid build script will not be readable after that.

also, I could not manage openssl as static lib, it is linked dynamically (while mbedtls is linked statically).
not sure what is better.

patches welcome here :)

.travis.yml
```diff
@@ -0,0 +1,103 @@
...
+
+env:
+  global:
+    - JOBS=3
+    - MBEDTLS_VERSION=2.2.1
+    - MBEDTLS_PREFIX=~/opt/mbed
+    - MBEDTLS_LIBS="-L$MBEDTLS_PREFIX/lib -lmbedtls -lmbedx509 -lmbedcrypto"
+    - MBEDTLS_CFLAGS="-I$MBEDTLS_PREFIX/include"
+    - OPENSSL_VERSION=1.0.1t
+    - OPENSSL_PREFIX=~/opt/ssl
```
mattock: This might be nitpicking, but why not _MBEDTLS_PREFIX=/opt/mbedtls_ and _OPENSSL_PREFIX=/opt/openssl_ instead?

mattock:
This PR seems to be a very reasonable starting point, so I'm inclined to ACK it. Any further comments from anyone?

.travis.yml
syzzer: As long as you mean ~/opt/{mbedtls,openssl}, I agree ;)

.travis.yml
```diff
@@ -0,0 +1,103 @@
...
+  - cd openssl-${OPENSSL_VERSION}/
+  - ./config shared --prefix=$OPENSSL_PREFIX -DPURIFY > build.log 2>&1 || (cat build.log && exit 1)
+  - make -j$JOBS > build.log 2>&1 || (cat build.log && exit 1)
+  - make install_sw > build.log 2>&1 || (cat build.log && exit 1)
+  - cd ..
+
+script:
+  - autoreconf -vi
+  - ./configure $CONFIG || (cat config.log && exit 1)
+  - make -j$JOBS > build.log 2>&1 || (cat build.log && exit 1)
```
syzzer: Hm, I agree on the usage of `> build.log 2>&1 || (cat build.log && exit 1)' in the 'install' bit, but this is the most interesting part of the build.  I'd rather have this output directly to travis.

syzzer:
Yes, this is a very reasonable starting point.  I'd be fine with merging this into master and improving it on the way.  Though in that case I'd recommend to just remove the failing XCode 6.4 builds for the time being.  (I contacted the mbedtls guys btw, they told me that they do not serve expired certs, so this must be some bug in the travis XCode 6.4 image).

chipitsine:
I'm not sure we which configuration are most interesting. I'm waiting for a feedback from you like

"let us start with "gcc+openssl+configure-options, clang+openssl+configure-options, ....."

current matrix is just an example, I picked random lines for configure, without any idea

.travis.yml
chipitsine: configure already outputs to stdout, 'cat config.log' means that if something went wrong, bring also config.log to analyse details

.travis.yml
mattock: Yeah, /opt would do, because extracting the tarballs will create the necessary subdirectories anyways.

chipitsine: currently I build both mbedtls and openssl in every build (it is cached by ccache).
if we will split single .travis.yml into several scripts in .travis folder, I think we can use conditional compilation (like building mbedtls only in appropriate builds)

putting those "if"s into .travis.yml pollute it, it is not readable after that

.travis.yml
mattock: The `command > build.log 2>&1  || cat build.log && exit 1` approach should work just fine. Everything is redirected to build.log, but if (and only if) the command return a non-zero exit code, then the entire output pushed to STDOUT.

Now the question is whether we want to see output from non-failing builds... the only use-case for that (afaics) is to be able to view build warnings. Is there any way we could get Travis-CI to notify us about build warnings without actually marking the build as failed?

.travis.yml
cron2: Hi,

On Tue, Jun 07, 2016 at 12:45:38AM -0700, Samuli Seppänen wrote:
> > +    - OPENSSL_VERSION=1.0.1t
> > +    - OPENSSL_PREFIX=~/opt/ssl
> 
> Yeah, /opt would do, because extracting the tarballs will create the necessary subdirectories anyways.
> 

~/opt is not /opt

gert
-- 
                                                           //www.muc.de/~gert/

.travis.yml
syzzer: Being able to see compiler warnings is enough reason on it's own.  Also, `2>&1` can mess up the order of log messages, making is harder to match stderr and stdout messages, and I don't see any reason to hide the make output for successful builds.

(And tbh, I think we _should_ fail the travis build if we get compiler warnings with the default build options, but that's maybe not for this pull request.)

syzzer:
Thanks for the updates.

I propose to start with the following build matrix:
linux + gcc + ./configure --with-crypto-library=openssl
linux + clang + ./configure --with-crypto-library=openssl
linux + gcc + ./configure --with-crypto-library=mbedtls
linux + clang + ./configure --with-crypto-library=mbedtls
osx + xcode 7 + ./configure --with-crypto-library=openssl
osx + xcode 7 + ./configure --with-crypto-library=mbedtls
linux + clang + ./configure --disable-crypto
linux + clang + ./configure --disable-lzo
linux + clang + ./configure --enable-small

That should cover a lot of common build problems.  We can always expand the matrix when we want more combinations tested.

chipitsine:
--with-crypto-library=openssl --disable-crypto  -->  https://travis-ci.org/chipitsine/openvpn/jobs/135816437

syzzer:
Hm, good point.  We should actually skip the failing tests.  I don't like 'allowed failure', because that way we'll probably won't notice real build failures.

More and more that votes for moving the scripts to e.g. .travis/install.sh and .travis/build.sh with more intelligent logic.

chipitsine:
I agree to move openssl/mbedtls build scripts to .travis
but I think we should not fail on "make check" with configure options, no matter we run it from travis or not

so, it makes sense to fix "make check" magic here

syzzer:
Yes, makes sense.  I agree.

chipitsine:
there're difference in openssl under mac os and linux, I beleive it is more simple to maintain separate scripts for each of them instead of polluting universal script with lots of "if"s

.travis/build-openssl-linux.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+if [ ! -f download-cache/openssl-${OPENSSL_VERSION}.tar.gz ]; then 
+	wget -O download-cache/openssl-${OPENSSL_VERSION}.tar.gz https://www.openssl.org/source/openssl-${OPENSSL_VERSION}.tar.gz;
```
syzzer: This will break every time a new OpenSSL version is released (because the referenced version is moved to old/).  Quite annoying, and out of our control. I'd rather not have to update this script for each openssl release.

The only way out I see is doing e.g. `git checkout --single-branch -b OpenSSL_1_0_2-stable', and then `git checkout $(git describe --abbrev=0 --tags)` to get to the most recent release...

chipitsine:
we run similar script for months https://github.com/OpenVPN/openvpn-gui/blob/master/.travis.yml#L30 without any issue yet.

also, I considered openssl as a git submodule here

thoughts ?

syzzer:
Ah, https://www.openssl.org/source/ apparently doesn't mimic ftp://ftp.openssl.org/source/, and does leave old links intact.  That's good, let's just download the tarbal then, that's more lightweight.

syzzer:
Wrt to the scripts: I think a couple of ifs is better than the code duplication we have now.  I have other things I must do now, but I can send a patch with a suggestion later.

chipitsine:
right now scripts are "almost duplicate", however, after some time spent, I will not wonder if each of them will differ a lot.

and I think it is more simple to change "just linux" script without being afraid to break something on mac os x.

ok, send your patch

.travis.yml
```diff
@@ -73,5 +73,5 @@ script:
   - autoreconf -vi
   - ./configure $CONFIG
   - make > build.log 2>&1 || (cat build.log && exit 1)
-  - ldd src/openvpn/openvpn
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then ldd src/openvpn/openvpn; fi
```
neuhalje: For MacOS we can use `otool -L` in lieu of `ldd`

chipitsine:
is it ok now ?

chipitsine:
knock-knock ?

syzzer:
As far as I'm concerned it is ready for review on the mailing list.

chipitsine:
@mattock , Samuli, can you please take care of mailing list, etc ?I'm not familiar with git-mail
and, as far as I see, I provided all the required patches here

cron2:
Hi,

On Thu, Aug 11, 2016 at 11:59:19AM -0700, Ilya Shipitsin wrote:
> @mattock , Samuli, can you please take care of mailing list, etc ?I'm not familiar with git-mail
> and, as far as I see, I provided all the required patches here

"all the required patches" sounds like "many" to me - and commits that
result from incrementally improving a patchset should not show up in the 
resulting commits.  

So, please merge to "as few patches as make sense" - and after that, 
sending to the mailing list is really easy

  git send-email --to=openvpn-devel@lists.sourceforge.net -5

(to send "the last 5 patches as one patch set")

gert

-- 
                                                           //www.muc.de/~gert/

mattock:
18 commits to create two files is definitely too much - one commit should be enough. One way to clean up the mess is to do something like this:

    git reset --soft <first-commit-id-before-the-patches>
    git status
    git add <modified files>
    git commit -s -m "travis-ci support"

Then you can force push your branch to GitHub:

    git push -f origin master

I did not test the commands, but I believe they should work. As for git-send-email: it is not too difficult to use, once you set it up correctly. Here are the basic instructions:

* https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project#_project_over_email

The section that starts with "You can also send the patches through an SMTP server" is the most relevant one. Basically you just add a [sendemail] section to ~/.gitconfig and you're done. You can test sending patches to yourself (--to=<my-email-address>), so that you don't accidentally send useless stuff to the list.

chipitsine:
I will make less commits soon.

chipitsine:
I squashed commits into 2 (mine and Steffan).
I added "Signed-Off-By" to reflect second commit ownership

as for the rest, I currently does not have a time to learn how to send git-email via GMail.

@mattock , can you please take care of the rest ? I remember, you promised me to do such kind of things.
Also, I'm not prepared to ask many times "what is going on" (because everyone is busy, remember, how long it took to review vlan patchset). It would be nice, if you would have take care of that as well.

mattock:
Now it looks good. The split into two patches makes sense, as the refactoring was done with syzzer. Configuring git-send-email for Gmail is not difficult:

* https://burzalodowa.wordpress.com/2013/10/05/how-to-send-patches-with-git-send-email/

Quoting the article:

    $ git config --global sendemail.smtpuser <your mail>
    $ git config --global sendemail.smtpserver smtp.googlemail.com
    $ git config --global sendemail.smtpencryption tls
    $ git config --global sendemail.smtpserverport 587
    $ git config --global sendemail.smtppass <your passwd>

Actual use of git-send-email has been covered above.

As said elsewhere, I can take care of sending patches for one-time contributors. Acting as a "patch proxy" for regular contributors such as you eats up my time, which I could use for genuine useful stuff instead. So please send the patches to the list yourself. The patches have already been reviewed, so it will almost certainly just get an ACK and get merged with minimal fuzz.

chipitsine:
@mattock , I'm not sure it will be merged. Remember how long did it take to merge VLAN patchset ? I think something is wrong in contribution process. So, can you take care of if it will take longer than a week ?

cron2:
Hi,

On Fri, Aug 26, 2016 at 08:44:50AM -0700, Ilya Shipitsin wrote:
> @mattock , I'm not sure it will be merged. Remember how long did it take to merge VLAN patchset ? I think something is wrong in contribution process. So, can you take care of if it will take longer than a week ?

Could you please stop complaining, and learn to use git send-email?

The VLAN patchset is LARGE and VERY intrusive, so it needs much focus to
review and merge - and it's a niche application, which only few people
have been asked about ever, and the trade-off in "who benefits" vs. "
who has to do the work" has not been in favour of that patchset so far
(it's still on my radar, but instead I waste my time argueing with people
that want to contribute but can't be bothered to contribute in a way
that is not creating *more* work for the core developers)

gert

-- 
                                                           //www.muc.de/~gert/

mattock:
@chipitsine : this PR has already been reviewed and ACKed, so once you git-send-email your changes will get merged shortly. Unlike you probably think, the use of git-send-email is not mandated just to annoy would-be contributors who would prefer to use GitHub. More on that below.

Each patch to OpenVPN _requires_ a proper review: there are many examples from past years where patches break things even when reviewed carefully. Without a review things would break more or less constantly, and the quality of the codebase would detoriate. For a very recent example, see commits https://github.com/OpenVPN/openvpn/commit/e9d64bc03742c96a3d7fe2a473c43d40e5ba2001 and https://github.com/OpenVPN/openvpn/commit/8cba9ffc57739eefa8e56b0738acd1dee2ff2396. Fortunately that particular problem was found in automated testing before shipping it in OpenVPN 2.3.12.

OpenVPN has dozens of esoteric options and use-cases, and it is extremely difficult to visualize all the corner-cases a change might affect. Our codebase is, as jamesyonan might put it, "entangled", which is why a lot of work has gone into refactoring it and making it easier to work with. This is one of the reasons we mandate the use of git-send-email - so that all active developers and power-users will at least know about a patch, and react to it _if_ it touches their particular use-case and the reviewer has managed to overlook something.

chipitsine:
ok, I sent 2 commits to mailing list

PR #55, opened 2016-06-22
testing: skip t_lpback.sh and t_cltsrv.sh if openvpn configured --disable-crypto

chipitsine:


syzzer:
This fixes the problem, in a more elegant way than I proposed, so I'm good with this.  But I do agree with Gert's remark on the mailing list: it would be even better if it printed something like 'skipping test'.  For consistency, the neatest way would be to somehow have the test framework report that, like it does now for t_client if no config is present.

chipitsine:
@mattock , can you please send the patch to list and take care of merge process ?

chipitsine:
up

mattock:
@chipitsine : please send the patch to the mailing list yourself. Helping out one-time contributors is useful, as they might just drop the ball if they would have to join the mailing list, learn git-send-email, etc. However, you're a regular contributor, and I don't want to proxy all of your current and future patches - it would be a lot of work for me, too.

PR #56, opened 2016-06-28
crypto: Add the --ca-pkcs11-id option

lkundrak:
Hi,

We're working on integrating NetworkManager with p11-kit's PKCS#11 remoting to allow using PKCS#11 with user initiated VPN (and Wi-Fi) connections: https://bugzilla.gnome.org/show_bug.cgi?id=767872

Currently OpenVPN seems to do pretty well with PKCS#11. That is, with an exception of CA handling. Users tend to need to supply their CA certificate and place it in their home directory. We'd like to use PKCS#11 instead because then we could avoid letting the VPN daemon being able to access files user's home directly (SELinux is already really unhappy about this) and would be able to use a nice certificate picker. Integration with smart cards would then be easier too.

I'm wondering if this and the patch makes sense to you, before I post it to the list.

Thank you,
Lubo

### DIFF
--- a/doc/openvpn.8
+++ b/doc/openvpn.8
@@ -4620,6 +4620,13 @@ by the standalone
 option.
 .\"*********************************************************
 .TP
+.B \-\-ca\-pkcs11\-id name
+Specify the serialized object id of a x509 CA certificate instead of a file
+certificate set with
+.B \-\-ca
+option.
+.\"*********************************************************
+.TP
 .B \-\-pkcs11\-id\-management
 Acquire PKCS#11 id from management interface. In this case a NEED-STR 'pkcs11-id-request'
 real-time message will be triggered, application may use pkcs11-id-count command to
--- a/src/openvpn/options.c
+++ b/src/openvpn/options.c
@@ -1710,6 +1710,7 @@ show_settings (const struct options *o)
   }
   SHOW_INT (pkcs11_pin_cache_period);
   SHOW_STR (pkcs11_id);
+  SHOW_STR (ca_pkcs11_id);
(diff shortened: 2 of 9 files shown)


syzzer:
At first sight, the option and the patch make sense (no time nor brains to do a proper review now, sorry).  One first nitpick/bike shed though, I'd prefer to have all pkcs11 options start with --pkcs11, so would suggest to rename it to --pkcs11-ca-id.

syzzer:
One related thing: won't a lot of other options also fail when openvpn can't access the homedir?  Shouldn't a user be able to run openvpn with config files in the homedir, or use something like --tls-auth (which people should really use!) or --log with files in their homedir?

cron2:
Hi,

On Tue, Jun 28, 2016 at 02:17:04PM -0700, syzzer wrote:
> One related thing: won't a lot of other options also fail when openvpn can't access the homedir?  Shouldn't a user be able to run openvpn with config files in the homedir, or use something like --tls-auth (which people should really use!) or --log with files in their homedir?

Well, you could always have the TLS stuff inline, and log to syslog...

All my openvpn configs do *not* read anything from the home dir, except
the initial .ovpn conf file...

gert

-- 
                                                           //www.muc.de/~gert/

syzzer:
... which is an excellent point:  you could also just inline the CA cert in the config, instead of specifying a path.  That should work fine right now.  (And if you decide to do so, please also support inlining of `--tls-auth`).

Still, something like `--pkcs11-ca-id` might make sense.

ordex:
No reply from the author for a bunch of time. Closing.

PR #59, opened 2016-07-22
distro: Update openvpn.init.d.rhel

emerlinsky:
Without this change `service openvpn status` command  exits with status 0 in situation when lock file and pid file exist but service is dead, which besides not complying with specs (http://refspecs.linuxbase.org/LSB_3.1.0/LSB-Core-generic/LSB-Core-generic/iniscrptact.html) prevents systems such as puppet from starting it back.

### DIFF
--- a/distro/rpm/openvpn.init.d.rhel
+++ b/distro/rpm/openvpn.init.d.rhel
@@ -229,6 +229,7 @@ case "$1" in
                 PID=$(cat $pidf)
                 if [ -e /proc/$PID ]; then
                     echo "openvpn is running. pid: $PID"
+                    exit 0
                 else
                     echo "openvpn is dead but lock file exists: $lock and pid file exists: $pidf"
                     exit 1
--- a/distro/rpm/openvpn.init.d.rhel
+++ b/distro/rpm/openvpn.init.d.rhel
@@ -224,17 +224,23 @@ case "$1" in
 	fi
 	;;
   status)
-	if [ -f $lock ]; then
-	    for pidf in `/bin/ls $piddir/*.pid 2>/dev/null`; do
-		if [ -s $pidf ]; then
-		    kill -USR2 `cat $pidf` >/dev/null 2>&1
-		fi
-	    done
-	    echo "Status written to /var/log/messages"
-	else
-	    echo "openvpn: service not started"
-	    exit 1
-	fi
+        if [ -f $lock ]; then
+            for pidf in `/bin/ls $piddir/*.pid 2>/dev/null`; do
+                PID=$(cat $pidf)
+                if [ -e /proc/$PID ]; then
+                    echo "openvpn is running. pid: $PID"
+                else
+                    echo "openvpn is dead but lock file exists: $lock and pid file exists: $pidf"
+                    exit 1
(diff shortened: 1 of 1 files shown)


distro/rpm/openvpn.init.d.rhel
```diff
@@ -224,17 +224,24 @@ case "$1" in
...
+            for pidf in `/bin/ls $piddir/*.pid 2>/dev/null`; do
+                PID=$(cat $pidf)
+                if [ -e /proc/$PID ]; then
+                    echo "openvpn is running. pid: $PID"
+                    exit 0
+                else
+                    echo "openvpn is dead but lock file exists: $lock and pid file exists: $pidf"
+                    exit 1
+                fi
+            done
```
mattock: So this part loops through all pidfiles in $piddir, but exits with 0 or 1, depending on the status of the first OpenVPN process /bin/ls returns. So, if there are several OpenVPN instances running, the return value of the "service status" depends on the status of the first OpenVPN instance, which I think is suboptimal.

That said, I don't know if that is the best we can do, or what is the convention in these cases. Thoughts?

emerlinsky: I understand your concern. But at the same time I hardly can imagine use
case when several openvpn instances are using same directory to store pid
files. From my experience,
such instances would be configured to use different service names and
different values for $piddir and few other variables for example by
creating a separate init script file to control each instance.

mattock: Throwing several OpenVPN config files into /etc/openvpn and having the init script launch them all seems to be a common practice/use-case on both RedHat and Debian derivatives; the support is there on CentOS 6 and Debian 7.  While I agree that having a separate init script for each connection would be cleaner, but without proper configuration management system (e.g. puppet, ansible, chef or salt) setting them up would be a pain.

Fortunately the init script has been superceded by per-connetion services (e.g. openvpn@work, openvpn@home) on CentOS 7 and Debian 8, which use systemd. Interestingly none of the distros seem to use _our_ unit files, and have invented their own, which all seem to differ.

Anyways, how about returning 0 if all OpenVPN connections are fine, and returning 1 if _any_ OpenVPN connection has failed? Does that sound reasonable?

dsommers: Having struggled with a few ScientificLinux 6  boxes over the years with several configs, I can understand this pain.  Generally speaking I have not used `service openvpn status` on my SL6/CentOS5 boxes precisely due to this issue. 

However, I would recommend not to do the exit immediately.  Loop through all configs, and report the status for each configuration file.  If one of the configs failed, then do `exit 1`.

By the way, I see this patch set uses exit 1, 2 and 3.  Are those standard exit values for upstart and/or traditional SysV init scripts?

And a final note.  Who will ensure these fixes gets included into the distro packages?  It seems every package manager likes to re-invent wheels (as we've seen with systemd too).  I'd like to change that though, but things needs to be followed up.  Otherwise we have nice init scripts in our source tree which nobody uses.

mattock:
We don't have this file anymore, so we can't patch it. Closing this PR.

PR #61, opened 2016-08-21
Treat the 'DOMAIN-SEARCH' option the same as 'DOMAIN'

yarmash:
Some servers push `dhcp-option DOMAIN-SEARCH` as well as `dhcp-option DOMAIN`. This change enables the script to understand `DOMAIN-SEARCH` and treat it the same as `DOMAIN`.

### DIFF
--- a/contrib/pull-resolv-conf/client.up
+++ b/contrib/pull-resolv-conf/client.up
@@ -60,11 +60,11 @@ while true; do
   [ -z "${fopt}" ] && break
 
   case ${fopt} in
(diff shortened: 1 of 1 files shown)


ordex:
without going into details, OpenVPN handles the two differently, therefore I am not sure why they should be made equal by the script.

ordex:
@yarmash do you have any extra explanation?

ordex:
Closing due to inactivity. Without a proper explanation, the change seems not useful.

PR #65, opened 2016-10-21
Add --bind-dev option.

BarbarossaTM:
This options allows the user to specify a network device the OpenVPN process
  should use when making a connection or binding to an address. This translates
  in setting the SO_BINDTODEVICE option to the corresponding socket (on Linux).

  When for example using VRFs on Linux [0] this allows making connections using
  the non-default VRF and having the tun/tap interface in the default VRF.

  It seems FreeBSD does not support the SO_BINDTODEVICE socket option, but has
  a similar one called IP_SENDIF. As I don't have any BSD running, this part is
  untested.

  Thanks to David Ahern (Cumulus Networks) for insights on this.

  [0] https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/networking/vrf.txt

Signed-off-by: Maximilian Wilhelm max@rfc2324.org

### DIFF
--- a/doc/man-sections/vrf.rst
+++ b/doc/man-sections/vrf.rst
@@ -1,0 +1,75 @@
+Virtual Routing and Forwarding
+---------------------------------------
+
+Options in this section relates to configuration of virtual routing and
+forwarding in combination with the underlying operating system.
+
+As of today this is only supported on Linux, a kernel >= 4.9 is recommended.
+
+This could come in handy when for example the external network should be only
+used as a means to connect to some VPN endpoints and all regular traffic
+should only be routed through any tunnel(s).
+This could be achieved by setting up a VRF and configuring the interface
+connected to the external network to be part of the VRF. The examples below
+will cover this setup.
+
+Another option would be to put the tun/tap interface into a VRF. This could
+be done by an up-script which uses the `ip link set` command shown below.
+
+
+**VRF setup with iproute2**
+
+Create VRF `vrf_external` and map it to routing table `1023`
+
+::
+
+      ip link add vrf_external type vrf table 1023
+
+Move `eth0` into `vrf_external`
+
+::
+
+      ip link set master vrf_external dev eth0
+
+Any prefixes configured on `eth0` will be moved from the `main` routing
+table into routing table `1023`
+
+
+**VRF setup with ifupdown**
+
+For Debian based Distributions `ifupdown2` provides an almost drop-in
+replacement for `ifupdown` including VRFs and other features.
+A configuration for an interface `eth0` being part of VRF `vrf_external`
+could look like this:
+
+::
+
+      auto eth0
+      iface eth0
+          address 192.0.2.42/24
+          address 2001:db8:08:15::42/64
+          gateway 192.0.2.1
+          gateway 2001:db8:08:15::1
+          vrf vrf_external
+
+      auto vrf_external
+      iface vrf_external
+          vrf-table 1023
+
+
+**OpenVPN config**
+
+--bind-dev device
+  Set the device to bind the server socket to to the VRF
+
+      --bind-dev vrf_external
+
+
+**Further reading**
+
+Wikipedia has nice page one VRFs: https://en.wikipedia.org/wiki/Virtual_routing_and_forwarding
+
+This talk from the Network Track of FrOSCon 2018 provides an overview about
+advanced layer 2 and layer 3 features of Linux
+ - Slides: https://www.slideshare.net/BarbarossaTM/l2l3-fr-fortgeschrittene-helle-und-dunkle-magie-im-linuxnetzwerkstack
+ - Video (german): https://media.ccc.de/v/froscon2018-2247-l2_l3_fur_fortgeschrittene_-_helle_und_dunkle_magie_im_linux-netzwerkstack
--- a/src/openvpn/init.c
+++ b/src/openvpn/init.c
@@ -3445,6 +3445,7 @@ do_init_socket_1(struct context *c, const int mode)
                             c->options.rcvbuf,
                             c->options.sndbuf,
                             c->options.mark,
+                            c->options.bind_dev,
                             &c->c2.server_poll_interval,
                             sockflags);
 }
--- a/src/openvpn/options.c
+++ b/src/openvpn/options.c
@@ -294,6 +294,9 @@ static const char usage_message[] =
 #if defined(TARGET_LINUX) && HAVE_DECL_SO_MARK
     "--mark value    : Mark encrypted packets being sent with value. The mark value\n"
     "                  can be matched in policy routing and packetfilter rules.\n"
+    "--bind-dev dev  : Bind to the given device when making connection to a peer or\n"
+    "                  listening for connections. This allows sending encrypted packets\n"
+    "                  via a VRF present on the system.\n"
 #endif
     "--txqueuelen n  : Set the tun/tap TX queue length to n (Linux only).\n"
 #ifdef ENABLE_MEMSTATS
@@ -6084,6 +6087,13 @@ add_option(struct options *options,
             }
         }
     }
+#ifdef TARGET_LINUX
+    else if (streq (p[0], "bind-dev") && p[1])
+    {
+        VERIFY_PERMISSION (OPT_P_SOCKFLAGS);
+        options->bind_dev = p[1];
+    }
+#endif
     else if (streq(p[0], "txqueuelen") && p[1] && !p[2])
     {
         VERIFY_PERMISSION(OPT_P_GENERAL);
--- a/src/openvpn/options.h
+++ b/src/openvpn/options.h
@@ -352,6 +352,7 @@ struct options
 
     /* mark value */
     int mark;
+    char *bind_dev;
 
     /* socket flags */
     unsigned int sockflags;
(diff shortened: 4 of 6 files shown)


BarbarossaTM:
This is an update to #64

spagu:
can this patch can be connected with this issue? https://forums.freebsd.org/threads/58019/

cron2:
Hi,

On Sun, Oct 23, 2016 at 07:00:56AM -0700, spagu wrote:
> can this patch can be connected with this issue? https://forums.freebsd.org/threads/58019/

Totally unrelated.

gert
-- 
                                                           //www.muc.de/~gert/

BarbarossaTM:
Hi @cron2 

I rebased the PR on v2.4.3 and reordered some function paramters are requested on the mailing list.

Please review and merge :-)

Best
Max

BarbarossaTM:
Hi again,

Just got feedback, that `--bind-dev` should only be shown on Linux as it's only evaluated on Linux, so I move the help text next to `--mark` which is already only shown on Linux systems and does something similar.

Please review and merge :-)

Best
Max

cron2:
Hi,

On Thu, Sep 21, 2017 at 11:19:30AM -0700, Maximilian Wilhelm wrote:
> Just got feedback, that `--bind-dev` should only be shown on Linux as it's only evaluated on Linux, so I move the help text next to `--mark` which is already only shown on Linux systems and does something similar.
> 
> Please review and merge :-)

I thought you have BSD code as well?

gert
-- 
                                                           //www.muc.de/~gert/

BarbarossaTM:
> I thought you have BSD code as well?

I had some in the earlier pull request which I found by googing. As I don' have any FreeBSD and zero experience with it, I can't test it. Any Arne suggested, that the code I googled doesn't work / the options don't exist. As there was no other feedback on *BSD, I dropped it.

Best
Max

cron2:
Hi,

On Thu, Sep 21, 2017 at 11:39:52AM -0700, Maximilian Wilhelm wrote:
> > I thought you have BSD code as well?
> 
> I had some in the earlier pull request which I found by googing. As I don' have any FreeBSD and zero experience with it, I can't test it. Any Arne suggested, that the code I googled doesn't work / the options don't exist. As there was no other feedback on *BSD, I dropped it.

Well, my initial requirement was "keep the things open so platform specific
stuff can be added later if we figure out how to do it on platform <x>",
and as such, moving inside #ifdef TARGET_LINUX is not a good approach.

gert

-- 
                                                           //www.muc.de/~gert/

BarbarossaTM:
> Well, my initial requirement was "keep the things open so platform specific stuff can be added later if we figure out how to do it on platform <x>", and as such, moving inside #ifdef TARGET_LINUX is not a good approach.

To the best of my knowledge the code can easily extended when someone figures out, how to do so on different platforms.

The rationale behind the use of the current ifdef is, that the option **now** only is shown on Linux, as **now** it's only supported on Linux. But that ifdef can easily be changed later.

What I initially referred to was https://lists.freebsd.org/pipermail/freebsd-net/2012-April/032064.html  which talks about IP_SENDIF, which according to https://wiki.freebsd.org/Networking seems to still exist. But as I said I have no FreeBSD (experience) and would rely on soeone who does, here.

As the initial code shows it **should** be rather easy to add a branch for FreeBSD here.

Max

lukastribus:
td;lr SO_BINDTODEVICE is not equivalent to "VRF selection", let's not imply that one way or the other; also VRF support and its APIs may never be portable.


SO_BINDTODEVICE is a linux-specific socket option supported since decades.
The VRF concept in linux was introduced in linux 4.3 - barely 2 years ago.

I want to emphasize that the SO_BINDTODEVICE option's primary use-case is NOT to assign sockets to VRF's. That works in 4.3 now too, but the primary use-case is to force an application to a specific interface (for DHCP for example).

So when BSD folks talk about implementing the SO_BINDTODEVICE option, we cannot assume they also mean VRF selection, just because that's the shortcut used in Linux. In fact, I don't even know if FreeBSD supports VRF's. It supports multiple FIBs and jails, just like Linux supported multiple FIBs and network namespaces long before VRF support. But whether FreeBSD supports VRF's just like Linux, I do not know.

What we know for sure is how to use VRF's in Linux, but another OS may implement a completely different API (based on a VRF name or id, instead of a interface name).

OpenBSD's ntpd.conf for example looks like this: `listen on address [rtable table-id]` [1].

How are we going to abstract that in openvpn? We can't, in my opinion.


So my suggestion is to not make this looks like a portable VRF selection feature. Let's call it what it is: an option to bind to a specific interface (currently supported on Linux).

If required for a commit, we can move all the logic out of the #ifdef TARGET_LINUX, and guard only the actual socket call with #ifdef TARGET_LINUX or SO_BINDTODEVICE, but we have to at least mention the fact that this currently only works in Linux (in the help text).


lukas

[1] https://man.openbsd.org/ntpd.conf

bao7uo:
Really hope this feature can be merged :)

dfawcus:
Yes it would be useful addition.

dsommers:
@bao7uo, @dfawcus Which feature do you expect SO_BINDTODEVICE or VRF binding?  As I understand the quite well written comment from @lukastribus, using SO_BINDTODEVICE for VRF binding is the wrong approach - it works now, but there is no guarantee for the future.  If the VRF binding is the expected outcome, then we need a different patch.

craig:
Please merge, we can add this feature for BSD when it becomes available there.

lukastribus:
> @bao7uo, @dfawcus Which feature do you expect SO_BINDTODEVICE or VRF binding? As I understand the quite well written comment from @lukastribus, using SO_BINDTODEVICE for VRF binding is the wrong approach - it works now, but there is no guarantee for the future.

I'm suggesting: just implement SO_BINDTODEVICE in openvpn, by advertising it exactly what the API is for: binding to a device (not binding to a VRF).

It's perfectly fine for people to then use it under Linux to select it for VRFs. It's not like the Linux behavior is going to change. However it is not and may never be portable across different OS.

schwabe:
So this also went up on the mailing list. The strlen+1 is off by one that needs to be fixed. And I am unsure how to test this feature or it does not work. Binding openvpn to eth0 and not having redirect-gateway enabled should make a working VPN connection but it doesn't.

dfawcus:
As to testing, on Linux one can set up an ECMP route to a target such that it will point to two interfaces.

Packets sent to that destination will experience some behaviour (possibly round-robin?)

Using a socket which has been bound to the specific interface will then ensure that packets to that destination from the socket are sent out of the desired interface, rather than experiencing the default kernel behaviour.

```
$ ip -d r sh
unicast 8.8.8.8 proto 11 scope global
        nexthop via 10.1.2.2 dev eth0 weight 1
        nexthop via 10.2.0.2 dev eth1 weight 2
```

Starbicycle:
This patch might solve a lot of my networking problems. Please add this feature to openvpn.

halvors:
Yes please, then i could avoid my hacky routing table move scripts on up/down :dancer:

eimann:
So after years of merging this finally and repackaging OpenVPN, when will this finally be merged?
Now with wg merged into Linux, I could stop wasting my time on this and move to iproute2/wg with netns which should achieve the same with a less hassle.

schwabe:
The current version of this patch was never send to the mailing list.

cron2:
So, this has taken WAY too long, and I apologize.  I wanted to really think long and hard about this, and how we could use this as building block on the way to a more generic VRF handling infrastructure - but that turned out to be "no time for it right now, think about it next week", and then it never happened.

Thanks to @lukastribus for clearly stating what is in here and what is not ("bind to device, which happens to allow some outer-VRF selection by implicit 'if the interface is in the VRF and we bind to it, the socket ends up in the VRF'").

I am not exactly happy with the way this is sprinkling code bits and structure bits all over the place, but I have to acknowledge that there might no "nicer" way to get the required information to the function where it is needed, at least not without rewriting all of socket.c... (and it makes sense to have these changes *not* all in a #ifdef LINUX conditional, because that would mess up readability even more).

So...
  - good enough (ACK)
  - please check Arne's comment wrt "strlen()+1" - what is the correct argument here, strlen() (not including 0-byte) or strlen()+1?  I don't know.
  - and then send to the openvpn-devel list, please, so I can pick it up from there.
  - if possible, add some notes how to test this most easily (what exactly does "bind to device" *do* on linux?  just force source IP address - and possibly VRF - or also routing?  Or do you need ip rule tables for that?)

BarbarossaTM:
Yay :-)

Aboue the "strlen()+1": This is the way to do it according to kernel documentation, so I'm keeping this. See https://github.com/torvalds/linux/blob/master/Documentation/networking/vrf.rst

Patch on the list coming up :-)

yanky83:
Yes, finally, thanks for getting this going! Been waiting for this for a while!

Let's make it happen please.

cron2:
code is in!

commit 19d3c602e7a3881cf7c2244b7c40b9958c0b7ebc (HEAD -> master, stable/master, mattock/master, gitlab/master, github/master)
Author: Maximilian Wilhelm <max@sdn.clinic>
Date:   Mon Jun 29 12:49:07 2020 +0200

    Add --bind-dev option.

documentation needs to be merged with the great RST rewriting of the manpage, but will not be forgotten.

Thanks for your patience and insistence!

BarbarossaTM:
I guess that's eventual consistency then :-)

Thanks for merging!

PR #66, opened 2016-11-06
 several travis-ci improvements

chipitsine:
```
(*) mingw build only configurations (no "make test")
(*) upgraded mbedtls to 2.4.0
(*) upgraded openssl to 1.0.1u
```

### DIFF
--- a/src/openvpn/options.c
+++ b/src/openvpn/options.c
@@ -4080,6 +4080,7 @@ usage(void)
     fprintf(fp, usage_message,
             title_string,
             o.ce.connect_retry_seconds,
+            o.ce.connect_retry_seconds_max,
             o.ce.local_port, o.ce.remote_port,
             TUN_MTU_DEFAULT, TAP_MTU_EXTRA_DEFAULT,
             o.verbosity);


.travis.yml
```diff
@@ -12,10 +12,11 @@ env:
   global:
     - JOBS=3
     - PREFIX="${HOME}/opt"
-    - MBEDTLS_VERSION="2.2.1"
+    - TAP_WINDOWS_VERSION=9.21.2
+    - MBEDTLS_VERSION="2.4.0"
     - MBEDTLS_CFLAGS="-I${PREFIX}/include"
     - MBEDTLS_LIBS="-L${PREFIX}/lib -lmbedtls -lmbedx509 -lmbedcrypto"
-    - OPENSSL_VERSION="1.0.1t"
+    - OPENSSL_VERSION="1.0.1u"
```
syzzer: We should probably change this to 1.0.2, 1.0.1 is going EOL on 2016-12-31.

.travis.yml
```diff
@@ -51,16 +52,18 @@ matrix:
       os: osx
       osx_image: xcode7.3
       compiler: clang
+    - env: CHOST=x86_64-w64-mingw32
+      os: linux
+      compiler: ": Win64 build only"
```
syzzer: How does this work / what does this do?

.travis/build-deps.sh
```diff
@@ -3,7 +3,7 @@ set -eux
 
 # Set defaults
 MBEDTLS_VERSION="${MBEDTLS_VERSION:-2.2.1}"
-OPENSSL_VERSION="${OPENSSL_VERION:-1.0.2h}"
+OPENSSL_VERSION="${OPENSSL_VERSION:-1.0.1u}"
```
syzzer: These seem to be duplicates of the globals in .travis.yml (probably my fault).  Should we remove these?

.travis/build-deps.sh
```diff
@@ -27,7 +27,7 @@ build_mbedtls () {
 
 download_openssl () {
     if [ ! -f "download-cache/openssl-${OPENSSL_VERSION}.tar.gz" ]; then
-        wget -P download-cache/ \
+        wget -P download-cache/ --no-check-certificate \
```
syzzer: --no-check-certificate?  why?

.travis/build-deps.sh
```diff
@@ -63,23 +76,41 @@ build_openssl () {
...
+      echo "deb http://archive.ubuntu.com/ubuntu xenial main universe" | sudo tee -a /etc/apt/sources.list.d/xenial.list
+      echo "deb http://archive.ubuntu.com/ubuntu xenial main" | sudo tee -a /etc/apt/sources.list.d/xenial.list
+      sudo apt-get update
+      sudo apt-get -y install dpkg mingw-w64
+      download_openssl
+      if [ ! -f "download-cache/tap-windows-${TAP_WINDOWS_VERSION}.zip" ]; then
+         wget -P download-cache/ http://build.openvpn.net/downloads/releases/tap-windows-${TAP_WINDOWS_VERSION}.zip
+      fi
+      unzip download-cache/tap-windows-${TAP_WINDOWS_VERSION}.zip
+      #cp tap-windows-${TAP_WINDOWS_VERSION}/include/tap-windows.h ${PREFIX}/${PREFIX}/include
```
syzzer: Commented out line should probably just be removed.

.travis/build-deps.sh
```diff
@@ -63,23 +76,41 @@ build_openssl () {
...
+      cp tap-windows-${TAP_WINDOWS_VERSION}/include/tap-windows.h include
```
syzzer: Instead of copying, it's cleaner to just add something like `-I$(pwd)/tap-windows-${TAP_WINDOWS_VERSION}/include/` to CFLAGS.

.travis/build-deps.sh
```diff
@@ -63,23 +76,41 @@ build_openssl () {
...
-if [ "${SSLLIB}" = "openssl" ]; then
-    download_openssl
-    build_openssl
-elif [ "${SSLLIB}" = "mbedtls" ]; then
-    download_mbedtls
-    build_mbedtls
+if [ ! -z ${CHOST+xxx} ]; then
+      # cross build
+      echo "deb http://archive.ubuntu.com/ubuntu xenial main universe" | sudo tee -a /etc/apt/sources.list.d/xenial.list
+      echo "deb http://archive.ubuntu.com/ubuntu xenial main" | sudo tee -a /etc/apt/sources.list.d/xenial.list
```
syzzer: Why do we need xenial packages?

.travis.yml
```diff
@@ -70,17 +73,27 @@ cache:
...
-  - if [ "${TRAVIS_OS_NAME}" = "osx" ]; then brew update     ; fi
-  - if [ "${TRAVIS_OS_NAME}" = "osx" ]; then brew install lzo; fi
+  - if [ "${TRAVIS_OS_NAME}" = "osx" ]; then
+      brew update;
+      brew install lzo;
+    fi
 
 install:
   - .travis/build-deps.sh > build-deps.log 2>&1 || (cat build-deps.log && exit 1)
+  - if [ ! -z "${CHOST+xxx}" ]; then unset CC; unset CXX; fi
```
syzzer: I haven't seen this `+xxx` before.  What does it do?  Is it some bash trick I should know?

.travis/build-deps.sh
chipitsine: for building openvpn master branch we need "xenial" mingw. after I added "xenial" repo, somehow it broke https downloads.

I haven't yet figured out why (and probably, we should just use --no-check-certificate until travis will introduce native support for xenial images, or ... migrate to docker)

.travis/build-deps.sh
chipitsine: openvpn master branch uses firewall manipulation (for dns filtering purposes), it does not compiled on trusty

.travis/build-deps.sh
chipitsine: yes

.travis/build-deps.sh
chipitsine: I couldn't get it running.

it would be nice if you will fork and try your idea

.travis/build-deps.sh
chipitsine: as for moving to 1.0.2, I guess the best is to perform tests on the same version of SSL, which is used for release build:

https://github.com/OpenVPN/openvpn-build/blob/master/generic/build.vars#L4

we should probably change 1.0.1 --> 1.0.2 everwhere

.travis.yml
chipitsine: it is common practice of running cross builds on travis-ci.
you just need to define "compiler" to something.

and, since travis uses $compiler --version, it is safe to define fake compilers in such way

.travis.yml
chipitsine: it is test "if variable CHOST is defined"

chipitsine: I do not mind to use more common approach here (if you will provide code)

PR #68, opened 2016-11-25
allow to enable plugin lookup with configure argument

eworm-de:
For plugin lookup (give relative path to plugin directory in configuration) we had to configure with something like this:
```
CFLAGS="$CFLAGS -DPLUGIN_LIBDIR=\\\"/usr/lib/openvpn/plugins\\\"" ./configure
```
This allows to pass `--enable-plugin-lookup` to configure to achieve the same. As a bonus we can be sure that install path and lookup path in openvpn binary are the same.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -303,7 +303,7 @@ AC_ARG_WITH(
 
 AC_ARG_WITH(
 	[plugindir],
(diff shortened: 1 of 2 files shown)


eworm-de:
Discussing on the mailing list... So closing here.

PR #69, opened 2016-11-28
Add redirect-gateway/redirect-private to script environment

AndersBlomdell:
While trying to get NetworkManager to obey the "push redirect-gateway/redirect-private" in the "--up"-script, I found that this was impossible since these commands were not visible in the environment. This is a first iteration trying to fix this. 

Since "redirect_gateway=<int>" was already part of the environment, the new environment variables are named "redirect_gateway_cmd/redirect_private_cmd"

### DIFF
--- a/src/openvpn/init.c
+++ b/src/openvpn/init.c
@@ -1181,6 +1181,34 @@ do_init_route_list (const struct options *options,
       /* copy routes to environment */
       setenv_routes (es, route_list);
     }
+  /* Inverse of redirect-* parsing in options.c */
+  if (options->routes->flags & RG_ENABLE) {
+    char *route_cmd;
+    char route_args[128];
+    if (options->routes->flags & RG_REROUTE_GW ||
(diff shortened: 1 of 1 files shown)


AndersBlomdell:
It might also make sense to change the code in options.c to validate that flags are compatible, e.g it's not clear from the documentation that the following directives are equivalent:

  "redirect-private ipv6" == "redirect-gateway !ipv4 ipv6"
  "redirect-private" == "redirect-gateway !ipv4"

ordex:
> It might also make sense to change the code in options.c to validate that flags are compatible, e.g it's not clear from the documentation that the following directives are equivalent:
> 
> "redirect-private ipv6" == "redirect-gateway !ipv4 ipv6" "redirect-private" == "redirect-gateway !ipv4"

I don't think the above are equivalent. How did you conclude so?

Other than that it seems you want to extend the environment variable just for debugging purposes. I don't think this is a good idea.

If still interested in having this change reviewed, please send your patch to the mailing list, so that a broader audience can comment and have a look. Thanks!

PR #70, opened 2016-12-04
Fix --disable-async-push configure option

doughdemon:


dsommers:
This should be fixed by now.  I already had a patch ready when noticing this PR.  This PR was not being seen earlier as we primarily use the openvpn-devel@lists.sourceforge.net mailing list for patch review and tracking.  But I ensured the commit would mention this PR.  For more detals, see commit e62eccf025aa60ec268787d2aa4a46310ed1cd60

PR #72, opened 2016-12-08
Fixed typos in Changes.rst

oittaa:


schwabe:
behaviour is just British english :P

mattock:
@schwabe : indeed. That said, there is one actual typo there. And we should probably try to conform to either british or american english, not mix the two randomly. Although, to be honest, most of our files probably have a mix of both.

oittaa:
@schwabe Yeah, but it was inconsistently used in the file.

selvanair:
Yes, inconsistent with two 'behaviour's and one 'behavior' :) I tend to write behaviour, but don't mind either.  Let's strive for correctness not consistency. As @mattock pointed out there are a couple of glaringly obvious typos fixed here, so the patch is good.

flichtenheld:
The actual typo has been fixed outside of this PR. Two remaining instances of "behaviour" but I don't think that is worth the PR. I would suggest just to close it...

PR #73, opened 2016-12-11
options: Remove IPv6 netmask limitation.

VittGam:
Hello,

I found that OpenVPN is refusing to configure an IPv6 netmask of less than 64 bits, or more than 124 bits.

This occurred to me in two different specific cases, one with a /128 netmask and another with a /56 netmask, when integrating OpenVPN into an already existing layer 3 network, where creating new subnets was not an option at all.

I had to workaround this apparently artificial limitation with `--iroute` on the servers, and a custom `--up` script on the clients. Then things started working as expected.

Maybe this check makes sense on Windows due to the many limitations of its network stack? But I think it definitely does not make sense on Linux.

So I'm proposing to remove this check.

What do you think?

Cheers,
Vittorio

### DIFF
--- a/src/openvpn/options.c
+++ b/src/openvpn/options.c
@@ -4730,12 +4730,6 @@ add_option (struct options *options,
       if ( get_ipv6_addr( p[1], NULL, &netbits, msglevel ) &&
            ipv6_addr_safe( p[2] ) )
         {
-	  if ( netbits < 64 || netbits > 124 )
-	    {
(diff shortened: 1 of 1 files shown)


cron2:
Hi.

We don't do pull requests, AND we keep what we have now - the code assumes subnets everywhere, and adding point-to-point configs would require extra code on some platforms, and lots of additional testing.

schwabe:
and yes if you have a want a point to point connection a real point to point with peer at both ends and explicit ifconfig-ipv6 should also work. but for the normal server to client mode we don't want to support this.

PR #74, opened 2016-12-15
Implement the passtos option for IPv6 packets and sockets.

VittGam:
Hello,

This pull request implements the `passtos` option for both IPv6 packets (parsing the ToS of IPv6 packets received inside the tunnel) and sockets (setting the ToS of IPv6 encrypted packets sent to the OpenVPN server).

Cheers,
Vittorio

### DIFF
--- a/src/openvpn/socket.h
+++ b/src/openvpn/socket.h
@@ -1102,12 +1102,26 @@ link_socket_write (struct link_socket *sock,
  * Extract TOS bits.  Assumes that ipbuf is a valid IPv4 packet.
  */
 static inline void
-link_socket_extract_tos (struct link_socket *ls, const struct buffer *ipbuf)
+link_socket_extract_tos_v4 (struct link_socket *ls, const struct buffer *ipbuf)
 {
   if (ls && ipbuf)
     {
       struct openvpn_iphdr *iph = (struct openvpn_iphdr *) BPTR (ipbuf);
       ls->ptos = iph->tos;
+      ls->ptos_defined = true;
+    }
+}
+
+/*
+ * Extract TOS bits.  Assumes that ipbuf is a valid IPv6 packet.
+ */
+static inline void
+link_socket_extract_tos_v6 (struct link_socket *ls, const struct buffer *ipbuf)
+{
+  if (ls && ipbuf)
+    {
+      struct openvpn_ipv6hdr *ip6h = (struct openvpn_ipv6hdr *) BPTR (ipbuf);
+      ls->ptos = ((ip6h->version_prio & 0x0f) << 4) | (ip6h->flow_lbl[0] >> 4);
       ls->ptos_defined = true;
     }
 }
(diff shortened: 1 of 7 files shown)


src/openvpn/syshead.h
```diff
@@ -396,7 +396,7 @@
 /*
  * Do we have the capability to support the --passtos option?
  */
-#if defined(IPPROTO_IP) && defined(IP_TOS) && defined(HAVE_SETSOCKOPT)
+#if defined(IPPROTO_IP) && defined(IP_TOS) && defined(IPPROTO_IPV6) && defined(IPV6_TCLASS) && defined(HAVE_SETSOCKOPT)
```
schwabe: this might need checking if that disables ipv4 tos on systems that have no ipv6 tos

src/openvpn/socket.h
```diff
@@ -1113,14 +1113,37 @@ link_socket_extract_tos (struct link_socket *ls, const struct buffer *ipbuf)
...
 /*
+ * Extract TOS bits.  Assumes that ipbuf is a valid IPv6 packet.
+ */
+static inline void
+link_socket_extract_tos_v6 (struct link_socket *ls, const struct buffer *ipbuf)
+{
+  if (ls && ipbuf)
+    {
+      struct openvpn_ipv6hdr *ip6h = (struct openvpn_ipv6hdr *) BPTR (ipbuf);
+      ls->ptos = ((ip6h->version_prio & 0x0f) << 4) | (ip6h->flow_lbl[0] >> 4);
```
schwabe: DId you check that this is endian safe?

schwabe:
Overall the change looks good. We would like to have all patches sent to the mailing list. See the Contributing file for more details.

src/openvpn/syshead.h
VittGam: I initially though the same thing, but then I had a look at 86e2fa5597fd1ad8e0102f134c63d6bc8cb7c291 and thought not to handle that case. Do you think that on some platforms, IPv6 ToS might not be supported while IPv6 is?

src/openvpn/socket.h
VittGam: It's handling a single byte, so it should be... Anyway I'm going to test this on an OpenWrt/LEDE big-endian router too.

VittGam:
Ok, I will.

src/openvpn/socket.h
schwabe: never say never:

struct iphdr {
#if defined(__LITTLE_ENDIAN_BITFIELD)
        __u8    ihl:4,
                version:4;
#elif defined (__BIG_ENDIAN_BITFIELD)
        __u8    version:4,
                ihl:4;
#else

cron2:
Hi,

On Thu, Dec 15, 2016 at 05:51:50AM -0800, Vittorio G (VittGam) wrote:
> +link_socket_extract_tos_v6 (struct link_socket *ls, const struct buffer *ipbuf)
> +{
> +  if (ls && ipbuf)
> +    {
> +      struct openvpn_ipv6hdr *ip6h = (struct openvpn_ipv6hdr *) BPTR (ipbuf);
> +      ls->ptos = ((ip6h->version_prio & 0x0f) << 4) | (ip6h->flow_lbl[0] >> 4);
> 
> It's handling a single byte, so it should be... Anyway I'm going to test this on an OpenWrt/LEDE big-endian router too.

I'm not sure if relevant: at this point, does the code know that the buffer
is indeed big enough, that is, "larger than sizeof(openvpn_ipv6hdr)"?

gert

-- 
                                                           //www.muc.de/~gert/

VittGam:
Yes and no: yes, because the is_ipv_X in proto.c checks that the packet is at least sizeof(struct openvpn_iphdr) (20 bytes) large, and so in this particular case it's kind of okay because the ToS is in the first two bytes of the IPv6 packet; no, because the is_ipv_X function should check against sizeof(struct openvpn_ipv6hdr) (40 bytes) for an IPv6 packet. So I guess that's going to be another patch! :)

ordex:
The "other patch" landed on the mailing list:

https://patchwork.openvpn.net/project/openvpn2/patch/20220917213154.29285-1-a@unstable.cc/

@VittGam any chance to send the patches to the ml?

ordex:
I took the freedom to rebase the patch and send it to the ml. This PR can be closed. Thanks!

PR #78, opened 2017-01-14
Fix user's group membership check in interactive service to work with domains

selvanair:
The patch ~will be~ is sent to the list  -- copying here for those who prefer to review PRs

Currently the username unqualified by the domain is used to validate
a user which fails for domain users. Instead authorize the user

(i) if the built-in admin group or ovpn_admin group is in the process token
(ii) else if the user's SID is in the built-in admin or ovpn_admin groups

The second check is needed to recognize dynamic updates to group membership
on the local machine that will not be reflected in the token.

These checks do not require connection to a domain controller and will
work even when user is logged in with cached credentials.

Resolves Trac: #810

v2: include the token check as described above

Signed-off-by: Selva Nair <selva.nair@gmail.com>

### DIFF
--- a/src/openvpnserv/interactive.c
+++ b/src/openvpnserv/interactive.c
@@ -1475,7 +1475,7 @@ RunOpenvpn(LPVOID p)
     }
 
     /* Check user is authorized or options are white-listed */
-    if (!IsAuthorizedUser(ovpn_user->User.Sid, &settings)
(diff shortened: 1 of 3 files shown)


cron2:
this can be closed now, right?

selvanair:
Yes, closing..

PR #79, opened 2017-01-17
Run tests needing 'ld --wrap' only where supported

neuhalje:
Run tests needing 'ld --wrap' only where supported

    --wrap is a feature of ld that is only available on Linux systems.
    http://www.mockator.com/projects/mockator/wiki/Wrap_Functions

    "Alas, this feature is *only available with the GNU tool chain on Linux*.
    GCC for Mac OS X does not offer the linker flag --wrap. "

    The implementation is still not optimal. Improvement would be an actual
    check against running with '--wrap'.

    Tested on Ubuntu precise (--wrap tests run) && Macos (--wrap tests not run).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -309,11 +309,17 @@ AC_ARG_WITH(
 )
 
 
+# Test if ld supports '--wrap' whcih currently is only supported on Linux
+# http://www.mockator.com/projects/mockator/wiki/Wrap_Functions
+AM_CONDITIONAL([SUPPORT_LD_WRAP], [ false ])
+
 AC_DEFINE_UNQUOTED([TARGET_ALIAS], ["${host}"], [A string representing our host])
 case "$host" in
 	*-*-linux*)
 		AC_DEFINE([TARGET_LINUX], [1], [Are we running on Linux?])
 		AC_DEFINE_UNQUOTED([TARGET_PREFIX], ["L"], [Target prefix])
+                # It might be better to actually run ld with --wrap
+                AM_CONDITIONAL([SUPPORT_LD_WRAP], [ true ])
 		;;
 	*-*-solaris*)
 		AC_DEFINE([TARGET_SOLARIS], [1], [Are we running on Solaris?])
--- a/tests/unit_tests/openvpn/Makefile.am
+++ b/tests/unit_tests/openvpn/Makefile.am
@@ -1,6 +1,9 @@
 AUTOMAKE_OPTIONS = foreign
 
-check_PROGRAMS = argv_testdriver buffer_testdriver
+check_PROGRAMS = 
+if SUPPORT_LD_WRAP
+check_PROGRAMS += argv_testdriver buffer_testdriver
+endif
 
 if ENABLE_CRYPTO
 check_PROGRAMS += tls_crypt_testdriver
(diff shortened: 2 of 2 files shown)


neuhalje:
Please would someone review && test this on their development machine?

configure.ac
```diff
@@ -309,11 +309,17 @@ AC_ARG_WITH(
...
+# http://www.mockator.com/projects/mockator/wiki/Wrap_Functions
+AM_CONDITIONAL([SUPPORT_LD_WRAP], [ false ])
+
 AC_DEFINE_UNQUOTED([TARGET_ALIAS], ["${host}"], [A string representing our host])
 case "$host" in
 	*-*-linux*)
 		AC_DEFINE([TARGET_LINUX], [1], [Are we running on Linux?])
 		AC_DEFINE_UNQUOTED([TARGET_PREFIX], ["L"], [Target prefix])
+                # It might be better to actually run ld with --wrap
+                AM_CONDITIONAL([SUPPORT_LD_WRAP], [ true ])
```
selvanair: Why not do an AC_RUN_IFELSE test to determine whether wrapping works or not -- that way we could handle for all platforms except cross-compiling which is anyway easy to handle as we cross-compile only for windows.

neuhalje: Because I frankly don't know how to do it. I spend an hour to google for a solution and then resorted to gather comments via this PR.

I would be more than happy if you could give me the correct `AC_RUN_IFELSE` invocation.

cron2:
Hi,

On Tue, Jan 17, 2017 at 04:58:35PM -0800, Selva Nair wrote:
> 
> >  AC_DEFINE_UNQUOTED([TARGET_ALIAS], ["${host}"], [A string representing our host])
>  case "$host" in
>  	*-*-linux*)
>  		AC_DEFINE([TARGET_LINUX], [1], [Are we running on Linux?])
>  		AC_DEFINE_UNQUOTED([TARGET_PREFIX], ["L"], [Target prefix])
> +                # It might be better to actually run ld with --wrap
> +                AM_CONDITIONAL([SUPPORT_LD_WRAP], [ true ])
> 
> Why not do an AC_RUN_IFELSE test to determine whether wrapping works or not -- that way we could handle for all platforms except cross-compiling which is anyway easy to handle as we cross-compile only for windows.

As Selva says.  It might not be straightforward, but the whole point of
having this autoconf madness is "test the system's capabilities" - otherwise
we could just get rid of it and do "#ifdef linux" in the code...

As far as I know, all our build environments *except* MacOS and mingw
(due to cross-building cmocka won't work, I think) have GNU ld (*today*),
so making this "linux only" is not the right way forward - and even
there, one distribution might decide to go with clang + llvm-ld, 
breaking --wrap...


I do understand that this is not trivially implemented :-) - what you
I'd do is go for AC_TRY_LINK, modifing LDFLAGS beforehand (add --wrap) 
and restoring it afterwards, and then trying to link something like
"real_printf()" or so.  We have two examples of AC_TRY_LINK in configure.ac 
today that could serve as a starting point.

(res_init() is one of the funky functions that are really macros on some
platforms, so "does it compile?" is not a suitable test there - thus,
AC_TRY_LINK)

gert
-- 
                                                           //www.muc.de/~gert/

selvanair:
On Wed, Jan 18, 2017 at 2:42 AM, Gert Doering <notifications@github.com>
wrote:

> I do understand that this is not trivially implemented :-) - what you
> I'd do is go for AC_TRY_LINK, modifing LDFLAGS beforehand (add --wrap)
> and restoring it afterwards, and then trying to link something like
> "real_printf()" or so. We have two examples of AC_TRY_LINK in configure.ac
> today that could serve as a starting point.
>


I opened a PR. Please take a look.

neuhalje:
Closed bc/ Selva send a way better patch

PR #80, opened 2017-01-18
Add a check for -Wl,--wrap support in linker

selvanair:
I used a dummy function but it may be better to use some thing like exit and __wrap_exit/__real_exit ? Any comments?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -582,7 +582,7 @@ AC_COMPILE_IFELSE(
 	[AC_MSG_RESULT([no])]
 )
 
-save_LDFLAGS="$LDFLAGS"
+saved_LDFLAGS="$LDFLAGS"
 LDFLAGS="$LDFLAGS -Wl,--wrap=exit"
 AC_MSG_CHECKING([linker supports --wrap])
 AC_LINK_IFELSE(
@@ -601,15 +601,11 @@ AC_LINK_IFELSE(
 	[
 		AC_MSG_RESULT([yes])
 		have_ld_wrap_support=yes
+		AC_DEFINE([HAVE_LD_WRAP_SUPPORT], [1], [Define to 1 if linker supports -Wl,--wrap option])
 	],
-	[	AC_MSG_RESULT([no])
-		have_ld_wrap_support=no
-	],
+	[AC_MSG_RESULT([no])],
 )
-LDFLAGS="$save_LDFLAGS"
-if test "$have_ld_wrap_support" = yes; then
-	AC_DEFINE([HAVE_LD_WRAP_SUPPORT], [1], [Define to 1 if linker supports -Wl,--wrap option])
-fi
+LDFLAGS="$saved_LDFLAGS"
 
 dnl We emulate signals in Windows
 AC_CHECK_DECLS(
--- a/configure.ac
+++ b/configure.ac
@@ -582,6 +582,35 @@ AC_COMPILE_IFELSE(
 	[AC_MSG_RESULT([no])]
 )
 
+save_LDFLAGS="$LDFLAGS"
+LDFLAGS="$LDFLAGS -Wl,--wrap=exit"
+AC_MSG_CHECKING([linker supports --wrap])
+AC_LINK_IFELSE(
+	[AC_LANG_PROGRAM(
+		[[
+			void exit(int);
+			void __real_exit(int);
+			void __wrap_exit(int i) {
+				__real_exit(i);
+			}
+		]],
+		[[
+			exit(0);
+		]]
+	)],
+	[
+		AC_MSG_RESULT([yes])
+		have_ld_wrap_support=yes
+	],
+	[	AC_MSG_RESULT([no])
+		have_ld_wrap_support=no
+	],
+)
+LDFLAGS="$save_LDFLAGS"
+if test "$have_ld_wrap_support" = yes; then
+	AC_DEFINE([HAVE_LD_WRAP_SUPPORT], [1], [Define to 1 if linker supports -Wl,--wrap option])
+fi
+
 dnl We emulate signals in Windows
 AC_CHECK_DECLS(
 	[SIGHUP],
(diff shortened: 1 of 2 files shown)


selvanair:
Would this be better for the AC_LINK_IFELSE test?
```
void exit(int);
void __wrap_exit(int i) {
    __real_exit(i);
}
```
Then in main just call `exit(0)`

cron2:
Hi,

On Wed, Jan 18, 2017 at 08:03:58AM -0800, Selva Nair wrote:
> Would this be better for the AC_LINK_IFELSE test?
> ```
> void exit(int);
> void __wrap_exit(int i) {
>     __real_exit(i);
> }
> ```
> Then in main just call `exit(0)`

Should work.

In case of a hypothetical linker that accepts --wrap but does "something
totally different", the code above would trigger an unresolved symbol
trying to link __real_exit() - and that would tell us "wrapping not
available".

Those that do not support it at all will fail on command line parsing
already.

gert
-- 
                                                           //www.muc.de/~gert/

selvanair:
Yes precisely that is the reasoning. Use of exit() instead of a randomly named dummy looks cleaner. Updated the PR as above.

If it looks good, will send to the list.

cron2:
Hi,

On Wed, Jan 18, 2017 at 05:21:21AM -0800, Selva Nair wrote:
> I used a dummy function but it may be better to use some thing like exit and __wrap_exit/__real_exit ? Any comments?

Generally speaking, I think we're getting close.

The actual configure.ac code confuses me - but maybe this is just my
lack of configure wisdom speaking.  

But anyway: why set "have_ld_wrap_support=yes/no" first, then use that 
to set AC_DEFINE([HAVE_LD_WRAP_SUPPORT]...) in an "if test" clause,
and *then* do AM_CONDITIONAL([HAVE_LD_WRAP_SUPPORT], ...)?  Are these
different things, even if called the same?

(ISTR that AC_DEFINE(...) goes to config.h?  Do we need that?)

gert
-- 
                                                           //www.muc.de/~gert/

selvanair:
AC_DEFINE is for config.h, AM_CONDITIONAL for using in makefiles. Currently we do not need the define in config.h but I can think of checking it in the source to provide an alternative to __wrap_xxx  -- I added it for "just in case". If its not needed we could directly set the AM_CONDITIONAL inside the test.

cron2:
Hi,

On Wed, Jan 18, 2017 at 11:30:17AM -0800, Selva Nair wrote:
> AC_DEFINE is for config.h, AM_CONDITIONAL for using in makefiles. Currently we do not need the define in config.h but I can think of checking it in the source to provide an alternative to __wrap_xxx  -- I added it for "just in case". If its not needed we could directly set the AM_CONDITIONAL inside the test.

Thanks for the explanation.  It might be useful (though the corresponding
makefile rule is going to be interesting, not calling "-Wl,--wrap=" if not
available so the code itself can work around it...).

I do see the beauty of having all AM_CONDITIONAL() calls nicely grouped
together, so if we still have all these other variables as "set a lowercase
thing first and later use that for AM_CONDITIONAL() an uppercase thing",
let's stick to it.

I think this is good for review on the -devel list - definitely a feature-ACK
from me, and the code looks like it should "do the right thing".  David
and Mathias Andree might have comments on "how to do/not do" configure.ac, so
I'd defer to their review before code-ACKing.

gert
-- 
                                                           //www.muc.de/~gert/

selvanair:
replying to cron2:
> In case of a hypothetical linker that accepts --wrap but does "something
> totally different", the code above would trigger an unresolved symbol
> trying to link __real_exit() - and that would tell us "wrapping not
> available".

In fact I found a system (omnios) where gnu ld is not the default but wrap is supported. Yet, just checking the linker option "-Wl,--wrap=exit" with an empty main() fails. Possibly same on other derivatives of opensolaris and solaris.

gcc -Wl,--wrap=exit on an empty main() program fails with __wrap_exit undefined! It seems this platform links exit() with every program and on wrapping exit it can't find it anymore. Our approach here does work correctly as we define __wrap_exit().

cron2:
Hi,

On Thu, Jan 19, 2017 at 09:34:38AM -0800, Selva Nair wrote:
> gcc -Wl,--wrap=exit on an empty main() program fails with __wrap_exit undefined! It seems this platform links exit() with every program and on wrapping exit it can't find it anymore. Our approach here does work correctly as we define __wrap_exit().

So we could just wrap something else, like "printf()"?  exit() might
be a bit too special.

As in, call the linker with --wrap=printf and the code would then
just call real_printf("hello, world\n"); - and if that succeeds in
linking, at least real_f()->f() wrapping works (and we bluntly assume
that this implies that f()->wrap_f() also works).

gert
-- 
                                                           //www.muc.de/~gert/

selvanair:
Hi,

On Thu, Jan 19, 2017 at 2:34 PM, Gert Doering <notifications@github.com>
wrote:

> On Thu, Jan 19, 2017 at 09:34:38AM -0800, Selva Nair wrote:
> > gcc -Wl,--wrap=exit on an empty main() program fails with __wrap_exit
> undefined! It seems this platform links exit() with every program and on
> wrapping exit it can't find it anymore. Our approach here does work
> correctly as we define __wrap_exit().
>
> So we could just wrap something else, like "printf()"? exit() might
> be a bit too special.
>
> As in, call the linker with --wrap=printf and the code would then
> just call real_printf("hello, world\n"); - and if that succeeds in
> linking, at least real_f()->f() wrapping works (and we bluntly assume
> that this implies that f()->wrap_f() also works).
>

It seems I was not being clear enough.. The current approach in the PR does
work, so no change required.

What I meant was naively checking that ld supports --wrap would have given
a false negative in opensolaris. We do define __wrap_exit() in the test
program so this approach does detect availability of wrap in solaris ld.

So all good.

Selva

P.S. This behaviour by solaris ld looked queer so thought of documenting it
in the PR conversation...

selvanair:
Patch committed. Closing.

PR #89, opened 2017-07-16
Support server cipher preference

c0r0n3r:


cron2:
Hi,

On Sun, Jul 16, 2017 at 01:23:21AM -0700, Szilárd Pfeiffer wrote:
> -- Commit Summary --
> 
>   * Support server cipher preference

Is there a good reason why one wouldn't want to enable this at all
times?

While the code is simple enough, adding yet another tls option is not
necessarily the best way forward - it will confuse people, cause extra
questions to be asked, and will eventually break things because the
testing requirement for "test with and without that option" just doubles.

gert


-- 
                                                           //www.muc.de/~gert/

pfeiffersz:
On 2017-07-16 17:17, Gert Doering wrote:
> Hi,
>
> On Sun, Jul 16, 2017 at 01:23:21AM -0700, Szilárd Pfeiffer wrote:
> > -- Commit Summary --
> >
> > * Support server cipher preference
>
> Is there a good reason why one wouldn't want to enable this at all
> times?
>
> While the code is simple enough, adding yet another tls option is not
> necessarily the best way forward - it will confuse people, cause extra
> questions to be asked, and will eventually break things because the
> testing requirement for "test with and without that option" just doubles.

First of all, I've just followed the method used by NGINX/Apache/... so
added an option.

Anyway, if there is a server configuration where tls-cipher is set
explicitly (by list of cipher suites), but in a bad order (for instance
RSA authentication is preferred over DHE), after the upgrade to a new
OpenVPN version, which forces the server's cipher preference order to
the clients, weaker cipher suite will be used than it had been used
before the upgrade.

Szilárd

ordex:
@schwabe is this something we may want to have? it seems meaningful, like we do with NCP (prefer server order).
However, I also vote for not having any extra option.

schwabe:
I thought we already had that in there but not configurable.

ordex:
> I thought we already had that in there but not configurable.

Quickly checked: yes, we do already set it:

```
ssl_openssl.c:323:#ifdef SSL_OP_CIPHER_SERVER_PREFERENCE
ssl_openssl.c:324:    sslopt |= SSL_OP_CIPHER_SERVER_PREFERENCE;
ssl_openssl.c-325-#endif
```

Therefore this PR is obsolete.

PR #91, opened 2017-08-28
Add match type subject-prefix to --verify-x509-name option

AntAgna:
Allows matching the beginning of the X509 subject in a similar manner as existing match type name-prefix.

Is useful to block access to all but a group of clients.


AntAgna:
The idea behind this is to be able to limit access to a subgroup of certificate holders.

Imagine this scenario :
Multiple VPN servers serve various purposes within an organization.
The same CA and client certificate files are used for the different VPN servers.
For one server, we want to limit access to a subset of the clients.

Here are the possible solutions I found :
Use ccd files and the --ccd-exclusive option to limit access.
Use "--verify-x509-name specific-org name" and "--x509-username-field OU" to accept only clients that have the proper OU.

The ccd method would work but it gets complex to manage for a large number of clients and it means configuring the VPN server every time a new user is added.

The verify-x509-name method works but x509-username-field causes a major problem. The x509-username-field option makes OpenVPN use that field as client identifier in the logs files, which makes all clients indistinguishable.


What I suggest is to add the match type "subject-prefix" to the --verify-x509-name option.
This would allow using this : --verify-x509-name 'C=..., ST=..., O=..., OU=specific-org' subject-prefix
To limit access to the clients that have the proper OU in their certificate without the need for the --x509-username-field option.

syzzer:
Hi,

The code looks nice and clean, but I don't think the use case is worth adding an option. Is there a reason why you don't use the a --tls-verify script for this?  That gives you a lot of flexibility and e.g. allows you to do stuff like prefix matching or much more fancy matching such as regex matching.

-Steffan

AntAgna:
I hadn't thought about using a --tls-verify script, I will try it.
But I think it would be better if it would be built in.

cron2:
Hi,

On Tue, Sep 26, 2017 at 03:50:34PM -0700, Antoine W. Campagna wrote:
> I hadn't thought about using a --tls-verify script, I will try it.
> But I think it would be better if it would be built in.

Too many special-case variants led us to what we have now - way over
100 special options, and a hard to maintain code base.

We aim for less options, not for more :-) - and special-case needs that 
can be done easily using one of the script hooks are usually better done
there.

gert

-- 
                                                           //www.muc.de/~gert/

chipitsine:
there's "contrib" folder for such examples.
@AntAgna , what do you think about better documentation (with examples) ?
I agree with Gert, tls-verify looks better and more flexible

AntAgna:
I have made it work using --tls-verify
[verify-client.py](https://github.com/OpenVPN/openvpn/files/1341322/verify-client.py.txt)

Thank you

AntAgna:
But I have to say though that the documentation about --tls-verify is not clear.
The documentation mentions that the script receives the "X509 common name (cn)".
But in fact, it receives the full subject.
And the formatting of the subject is not mentioned.
Is it using commas , or slashes / ?
Fortunately, my method works with both.

syzzer:
Yeah, the --tls-verify docs could use some improvement. It's on my list of things to update some time, but so are 1001 other things :(  But documentation patches are welcome ofc ;-)

Zrin:
There are several reasons for introduction of this option and against use of --tls-verify:

- readability: this option clearly states what it does in the server configuration - givesbetter maintainability

- simplicity: prevent introduction of point of failures

- performance: invoking a script (process) for every TLS negotiation is a significant performance hit; consider also DoS attacks

- it's logical: to creates a sub-CA for networking / VPN and restrict VPN clients in that way. It's logical to be able to match a part of the DN, this is actually the reason why DN is structured as it is - to be able to match parts of it.

Please reconsider adding match type 'subject-prefix' to --verify-x509-name.

Thank you!

PR #96, opened 2017-12-03
Add common_name to the conv method

mkarvan:
Add common_name to the conv method.
This allows the common_name to be accessible in PAM.

### DIFF
--- a/src/plugins/auth-pam/auth-pam.c
+++ b/src/plugins/auth-pam/auth-pam.c
@@ -637,9 +637,16 @@ my_conv(int n, const struct pam_message **msg_array,
                         ret = PAM_CONV_ERR;
                     }
                     break;
+                
+                case PAM_TEXT_INFO:
+                    aresp[i].resp = strdup(up->common_name);
+                    if (aresp[i].resp == NULL)
+                    {
+                        ret = PAM_CONV_ERR;
+                    }
+                    break;
 
                 case PAM_ERROR_MSG:
-                case PAM_TEXT_INFO:
                     break;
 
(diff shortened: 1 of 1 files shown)


src/plugins/auth-pam/auth-pam.c
```diff
@@ -637,9 +637,16 @@ my_conv(int n, const struct pam_message **msg_array,
                         ret = PAM_CONV_ERR;
                     }
                     break;
+                
+                case PAM_TEXT_INFO:
+                    aresp[i].resp = strdup(up->common_name);
+                    if (aresp[i].resp == NULL)
```
ordex, line 643: we normally prefer "if (!something)"

ordex:
I am not a PAM expert but it seems to make sense. Can you please this patch to the mailing list with a proper comment and SoB line?

Thanks!

ordex:
Actually I did it for you.

Killed some trailing spaces and added SoB line.

Discussion can continue on the mailing list.

selvanair:
In case anyone stumbles upon this in future:

> Add common_name to the conv method. This allows the common_name to be accessible in PAM.

common-name is already accessible via PAM. Have your PAM module prompt for it and define the prompt string in the plugin argument using COMMONNAME as the string substitution keyword.

E.g., if PAM prompts start with "user", "password" and "common-name", in the config file one would have 

`plugin /path/to/openvpn-auth-pam.so "my-pam-confg user USERNAME password PASSWORD common-name COMMONNAME"`

where "my-pam-config" is the name of the PAM config file to be used.

PR #97, opened 2017-12-05
Add a warning for disabled DHCP media sense on Window

engycz:
When DHCP media sense configuration is disabled, network applications
including DHCP client will not receive information about link status
changes and the link seems to be always connected. This lead to the
non-renewal DHCP address on OpenVPN connect.

DHCP media sense status can by shown with command
"netsh interface ipv4 show global"

There are several reports of problems with DHCP address renewal.
https://community.openvpn.net/openvpn/ticket/665
https://community.openvpn.net/openvpn/ticket/807

Added checking of disabled DHCP media sense and print a warning with
forced dhcp-renew option and suggestion to enable DHCP media sense.

Signed-off-by: Jiří Engelthaler <EngyCZ@gmail.com>

### DIFF
--- a/src/openvpn/tun.c
+++ b/src/openvpn/tun.c
@@ -3828,6 +3828,46 @@ get_panel_reg(struct gc_arena *gc)
 }
 
 /*
+ * Return DhcpMediaSense enabled value
+ */
+static bool
+get_dhcp_media_sense(void)
+{
+    const char registry_tcpip_params[] = "SYSTEM\\CurrentControlSet\\Services\\Tcpip\\Parameters";
+    const char disable_dhcp_media_sense_string[] = "DisableDHCPMediaSense";
+    HKEY tcpip_params;
+    LONG status;
+    DWORD len;
+    DWORD disable_dhcp_media_sense;
+    DWORD data_type;
+    bool ret_value = true;
+
+    status = RegOpenKeyEx(HKEY_LOCAL_MACHINE, registry_tcpip_params, 0, KEY_READ, &tcpip_params);
+    
+    if (status != ERROR_SUCCESS)
+    {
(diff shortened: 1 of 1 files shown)


basezen:
This report helped me perfectly, but the warning was never printed in my client-side log. I do receive other warnings. Verbosity level is 3. Can we re-open this to see why I'm not getting the warning? This is Windows Server 2012 w/latest updates and latest OpenVPN 2.4.5 on client and server.

Spacefish:
Would be nice to have this merged.. Windows Server 2019 seems to have DhcpMediaSense disabled by default and we run into this problem!

cron2:
Hi,

On Sun, Jul 07, 2019 at 04:22:50PM -0700, Spacefish wrote:
> Would be nice to have this merged.. Windows Server 2019 seems to have DhcpMediaSense disabled by default and we run into this problem!

The place to get this done is on the openvpn-devel mailing list.

A v2 patch with the suggested changes, rebased to master, needs to be
sent to the list for this to proceed.

gert
-- 

ordex:
@selvanair do you think it is still meaningful to purse the solution provided in this PR? Or should we go with "dhcp_new" is always on? (saw your comments in https://community.openvpn.net/openvpn/ticket/665)

selvanair:
As argued in ticket 665 that you point out, I still believe its so much simpler to just always enable dhcp-renew when dhcp is in use. The alternative of throwing more code at settings like "dhcp-media-sense" is engaging in a never ending fight with Windows internals.

ordex:
Ok, then we should go with the approach suggested by Selva and avoid further corner cases.
This patch as is therefore obsolete.

If interested, please send a patch to the ml that enables dhcp-renew everytime dhcp is used.

PR #98, opened 2017-12-17
README.down-root: Fix plugin module name

scop:
The module name is openvpn-plugin-down-root.so, not openvpn-down-root.so.


flichtenheld:
As part of an initiative to clean up the GitHub PR submissions, sending this PR to the mailinglist for inclusion.

cron2:
merged

PR #99, opened 2018-01-01
Support naked CN subjects like 'CN=example.com'

sbrudenell:
I've been using LetsEncrypt for my certificates, with a whitelist of CNs using the `verify-cn` sample script.

LetsEncrypt issues certificates with subjects which include only a CN, like `CN=www.brudenell.name`. The script's current regex misses these.

Unfortunately I don't intend to subscribe to `openvpn-devel@lists.sourceforge.net` as seems to be necessary to contribute. Sourceforge has a history of exploiting users, and per their signup page where I must opt out of "newsletters" and "exclusive offers", I don't trust they've changed.

Please close this PR and/or incorporate the change, as you will.

### DIFF
--- a/sample/sample-scripts/verify-cn
+++ b/sample/sample-scripts/verify-cn
@@ -37,8 +37,8 @@ if ($depth == 0) {
     # If so, parse out the common name substring in
     # the X509 subject string.
 
-    if ($x509 =~ / CN=([^,]+)/) {
-        $cn = $1;
+    if ($x509 =~ /( |^)CN=([^,]+)/) {
+        $cn = $2;
 	# Accept the connection if the X509 common name
 	# string matches the passed cn argument.
 	open(FH, '<', $cnfile) or exit 1; # can't open, nobody authenticates!


syzzer:
Patch looks good, but why trust an external party to man-in-the-middle your VPN? You're probably much better off creating your own CA for openvpn.  See e.g. easy-rsa for simple tooling to do that.

sbrudenell:
I suppose one could debate whether LetsEncrypt (or any third party CA) is trustworthy for any purpose.

Ultimately, my answer is that I trust LetsEncrypt not to screw me, before I trust myself. Even easy-rsa is tedious for personal / homelab use. I've had trouble keeping OpenVPN consistently configured because of this, and fallen back to bad habits like exposing ssh to the Internet to work around it. My strategy is to make OpenVPN low-maintenance for myself, so it becomes a given rather than an ideal. I know this is my personal failing as a sysadmin, but 10+ years of (poorly) using OpenVPN has shown I'm not going to change.

You do make a good point though. Perhaps the ideal for my case is a self-hosted ACME responder, to make certificate management more automatic but not rely on third parties.

sbrudenell:
Thoughts?

syzzer:
I think the patch is fine.  If you submit it to openvpn-devel@lists.sourceforge.net using git send-email, it will likely be applied.  I'm closing the merge request just to be clear we don't accept patches through github.

PR #104, opened 2018-03-11
systemd: Enable systemd's auto-restart feature for client profiles.

d1b:
Signed-off-by: David <db@d1b.org>

### DIFF
--- a/distro/systemd/openvpn-client@.service.in
+++ b/distro/systemd/openvpn-client@.service.in
@@ -18,6 +18,8 @@ DeviceAllow=/dev/net/tun rw
 ProtectSystem=true
 ProtectHome=true
 KillMode=process
+RestartSec=5s
+Restart=on-failure
 
 [Install]
 WantedBy=multi-user.target


d1b:
Related: https://github.com/OpenVPN/openvpn/commit/a4686e99b047081f0ef6f7945450183088464aa5 .

dsommers:
Currently we are not quite decided if we want to enable this by default on clients for all configurations.  We recognise this is quite useful for many setups, but all of them?  That decision was easier on the server side, the vast majority of sys-admins we've heard from liked the idea on the server side - a server is of little use if the services it wants to provide is unavailable.

A client (host) might have multiple client configurations, with different expectations.  So it is harder to have a default policy in this context.  And especially when it is fairly easy to add this manually, even on a per-configuration scenario:

    # systemctl edit openvpn-client@CONFIGNAME

And then add the `RestartSec=` and `Restart=` settings inside a `[Service]` section.

We currently believe this approach is better for the vast majority of users.

d1b:
> Currently we are not quite decided if we want to enable this by default on clients for all configurations. We recognise this is quite useful for many setups, but all of them
...
> A client (host) might have multiple client configurations, with different expectations. So it is harder to have a default policy in this context. And especially when it is fairly easy to add this manually, even on a per-configuration scenario:

Sure but the condition being changed / modified here is on failure. I do understand the difficulty with having the client reconnect though. One example I was thinking about is if an organisation doesn't want 
their users accidentally connecting to the vpn while in the office, then preventing the systemd service from retrying could/would be useful. (However, I would expect a fair mount of those users to be using a graphical client such as `network-manager` on linux to connect to the vpn).

> We currently believe this approach is better for the vast majority of users.

@dsommers sure. Perhaps the answer here is to document the default behavior and how to change it.

dsommers:
Yeah, improving the documentation is definitely a good way forward regardless.  So I'm willing to review and accept patches for this topic, and we can see where this discussion goes further.

But please remember: patches needs to end up in the public mailing list in the end.  We can discuss things here via pull-requests resulting in a final patch, but the final result needs to end up on the mailing list.

vermeeren:
Today I was bitten by this issue. I adjusted some seemingly irrelevant options on a OpenVPN server instance and restarted it, expecting the clients to handle it gracefully. Because the clients drop a bunch of privileges however *and* OpenVPN decided recreation of the `tun` device was needed it ended in an explosion where all clients fatal exited and as a result the `openvpn-client@.service` for them failed. Quite a mess I had to manually untangle to get critical systems talking to each other again.

Personally I think the default should be to retry. A system service should do it's best to keep working, as it usually means it's part of a server or managed workstation, etc, and not a human user interacting with a VPN GUI or similar. Especially nasty if you rely on the VPN for remote access to the other machines, luckily didn't happen for me today.

I genuinely don't see a use case where you *don't* want it to automatically restart and try again on failure! Imagine if a DHCP client would stop trying if in encountered a very weird situation and fatal exited, resulting in a host without network connectivity. This with OpenVPN client feels very similar in principle to that.

I'll try sending a mail to the mailing list referring to this PR, hopefully this can be updated quite soon for future releases.

cc @dsommers

PR #107, opened 2018-07-20
list.c - fix for use of sizeof() on a pointer type in hash_func

TimRots:
fix will ensure sizeof() returns the size of the data structure itself, and not the size of the pointer to the data structure. (CWE-467).

### DIFF
--- a/src/openvpn/list.c
+++ b/src/openvpn/list.c
@@ -225,7 +225,7 @@ hash_remove_marked(struct hash *hash, struct hash_bucket *bucket)
 uint32_t
 void_ptr_hash_function(const void *key, uint32_t iv)
 {
-    return hash_func((const void *)&key, sizeof(key), iv);
+    return hash_func((const void *)&key, sizeof(*key), iv);
 }
 
 bool


selvanair:
Only thing one can say about this is that the intent of the original code is unclear especially since it appears to be unused. But how do you know the intention was not to take size of the pointer? 

In any case dereferencing a void pointer is almost always a bug -- don't do that. Its undefined in C although gcc will happily evaluate it to 1 byte (unl The pointer is void, so unless you cast it back to what it really is, you can't find the size of what it points to.
k
That said, just deleting this function may not be a bad idea -- it looks unused, contains a superfluous cast and looks useless.

TimRots:
Thank you for caring and taking time to give your thoughts on this PR.

I agree and see now that removing the function entirely would be the best solution for the issue on-hand, and will propose a patch for this to the mailing-list after some more thorough review of this code base.

selvanair:
> I agree and see now that removing the function entirely would be the best solution for the issue on-hand,

To nit-pick, there is no "issue" in the present code. You are just imagining that there is a bug. But, yes, it looks like an unused function with a questionable intent.

cron2:
Hi,

On Fri, Jul 20, 2018 at 03:57:24PM -0700, Tim Rots wrote:
> fix will ensure sizeof() returns the size of the data structure itself, and not the size of the pointer to the data structure. (CWE-467).
> 
>   https://github.com/OpenVPN/openvpn/pull/107
> 
> -- Commit Summary --
> 
>   * list.c - fix for use of sizeof() on a pointer type in hash_func

That fix is actually *wrong*.

What the code does is "get the address of the pointer" and "the size
of the pointer", and use that as input into the hash function.

Whether or that is a useful thing to do is indeed questionable (as
getting the address for a local variable will be "something on the 
stack", and I wonder why you'd want to hash that) - but technically it
is *correct*.  It even says so, "void_ptr_hash_function"...

Changing the sizeof part to "what the pointer points to" is changing
the function semantics - effectively reducing the hashed data space 
from 4 (or 8) to "1".  Not fixing anything.


Arguably we could get rid of it in git master, as it's not used
(as is void_ptr_compare_function()), but then, it doesn't do harm 
and there are more interesting areas to clean up...  so I wouldn't put 
this on high priority.

gert
-- 

PR #109, opened 2018-10-07
man: correct a --redirection-gateway option flag

samy-mahmoudi:
Replace "servers" with "peers" in the description
of the --redirection-gateway option flag local.

Patch submitted to openvpn-devel.


flichtenheld:
This PR has been already been applied in Git (commit f6bac113bcde4e342caf16d88e0a3a8e71085c90) and can be closed

PR #116, opened 2018-12-20
distro: Fix error message when using RHEL init script

RichieB2B:
In RHEL 7 `/etc/sysconfig/network` is no longer used (still there but empty). This results in the following error when openvpn starts:
```
Dec 20 09:01:25 localhost openvpn: /etc/rc.d/init.d/openvpn: line 94: [: =: unary operator expected
```


mattock:
We don't have this file anymore, so it's not possible to patch it. Closing this PR.

PR #121, opened 2019-02-18
Do not set pkcs11-helper "safe fork mode"

hillu:
From the pkcs11-helper API documentation about pkcs11h_setForkMode():

> This funciton is releavant if PKCS11H_FEATURE_MASK_THREADING is
> set. If safe mode is on, the child process can use the loaded
> PKCS#11 providers but it cannot use fork(), while it is in one of
> the hooks functions, since locked mutexes cannot be released.

As far as I can tell, pkcs11-helper functionality is not used in a
child process that is created after initialization. Even if OpenVPN is
turned into a daemon, the pkcs11-helper library is only initialized
after calling possibly_become_daemon(), i.e. in the child process. All
other uses of fork() are immediately followed by an exec()

This simple change fixes the symptoms described in both
<https://community.openvpn.net/openvpn/ticket/538> (hang on password
prompt when systemd support is enabled) and
<https://community.openvpn.net/openvpn/ticket/1157> (hang on
initialization with newer versions of pkcs11-helper).

I have successfully tested that this makes the described symptoms go
away. For this, I used a YubiKey NEO on Debian/stable, a rebuild of
OpenVPN 2.4.6 and two versions of libpkcs11-helper:

- libpkcs11-helper 1.21-1 from Debian/stretch
- a backport of libpkcs11-helper 1.25-1 from Debian/buster

### DIFF
--- a/src/openvpn/pkcs11.c
+++ b/src/openvpn/pkcs11.c
@@ -312,7 +312,7 @@ pkcs11_initialize(
 
     pkcs11h_setLogLevel(_pkcs11_msg_openvpn2pkcs11(get_debug_level()));
 
-    if ((rv = pkcs11h_setForkMode(TRUE)) != CKR_OK)
+    if ((rv = pkcs11h_setForkMode(FALSE)) != CKR_OK)
     {
         msg(M_FATAL, "PKCS#11: Cannot set fork mode %ld-'%s'", rv, pkcs11h_getMessage(rv));
         goto cleanup;


src/openvpn/pkcs11.c
```diff
@@ -312,7 +312,7 @@ pkcs11_initialize(
 
     pkcs11h_setLogLevel(_pkcs11_msg_openvpn2pkcs11(get_debug_level()));
 
-    if ((rv = pkcs11h_setForkMode(TRUE)) != CKR_OK)
+    if ((rv = pkcs11h_setForkMode(FALSE)) != CKR_OK)
```
alonbl: as far as I remember, openvpn has early initialization then fork as daemon then runs. I expect the first phase to run with safe fork and then only after daemonizing to set unsafe fork.

syzzer: The `possibly_become_deamon()` is call is before the call to `pkcs11_initialize()` (which is in `context_init_1()`. This makes me think the current patch should be fine.

hillu: Even if `possibly_become_daemon()` was called after initializing the /pkcs11-helper/ library, I believe that setting "safe fork mode" would not be needed because its functionality is never used from multiple processes/threads.

alonbl:
On Thu, May 2, 2019 at 6:19 PM Hilko Bengen <notifications@github.com>
wrote:

> *@hillu* commented on this pull request.
> ------------------------------
>
> In src/openvpn/pkcs11.c
> <https://github.com/OpenVPN/openvpn/pull/121#discussion_r280468831>:
>
> > @@ -312,7 +312,7 @@ pkcs11_initialize(
>
>      pkcs11h_setLogLevel(_pkcs11_msg_openvpn2pkcs11(get_debug_level()));
>
> -    if ((rv = pkcs11h_setForkMode(TRUE)) != CKR_OK)
> +    if ((rv = pkcs11h_setForkMode(FALSE)) != CKR_OK)
>
> Even if possibly_become_daemon() was called after initializing the
> /pkcs11-helper/ library, I believe that setting "safe fork mode" would not
> be needed because its functionality is never used from multiple
> processes/threads.
>

Please verify if slotevents are in used or not, if they are, must use safe
fork if fork is being used post initialization.

And as far as I remember, openvpn has a late fork mechanism so that PIN is
queried before going daemon, please look closely on the code, it is not
trivial.

hillu:
@alonbl To me it looks as if there's no use of slotevent APIs. The only thing I am not sure about is the use of the `PKCS11H_SLOTEVENT_METHOD_AUTO` constant around https://github.com/OpenVPN/openvpn/blob/4ded2deda0d6d33d38fbbd1f4cd4e9acc0836a89/src/openvpn/pkcs11.c#L400

syzzer:
So, coming back to this change again.

Thanks @alonbl for providing useful pointers.

The initialization order has been changed around 2015. Since then, OpenVPN initializes the crypto - including pkcs11 - always after daemonizing. Any PIN / password is always queried before that point. Since this patch will only be applied to release/2.4 and master, we're good in that aspect.

With respect to slot events, I believe we're good too. If I understand the code correctly (@alonbl, please correct me if I'm wrong), slot events are only used if someone calls pkcs11h_setSlotEventHook(), which OpenVPN doesn't do.

Summarizing, unless @alonbl tells me I'm wrong I think this patch is good and should be merged.

alonbl:
On Sat, Jul 20, 2019 at 1:07 PM syzzer <notifications@github.com> wrote:

> Summarizing, unless @alonbl <https://github.com/alonbl> tells me I'm
> wrong I think this patch is good and should be merged.


You are good to go, although you may have a leak of handles from `PKCS#11`
provider into child processes. Per PKCS#11 spec after each fork if
`PKCS#11` is not used, the C_Finalize must be called. This can be done by
terminating the pkcs11-helper after any fork, however, this will also
attempt to lock, I should probably add a new method that terminates
pkcs11-helper without locking to be called in these cases.

cron2:
commit 59e45a8bbc9084329c683730325bc5b676058e71 (master)
commit 59e45a8bbc9084329c683730325bc5b676058e71 (release/2.4)

PR #128, opened 2019-08-15
Changes open vpn exit code defines

Kutashov:
It's better to have distinct enums for more straightforward debug of exit function.


ordex:
@cron2 what do you think about this? It makes sense to me. As long as result is non-zero, assumption is that this is a failure.

ordex:
Actually I have sent this patch to the ml, so the PR can be closed. The discussion can continue on the mailing list.

PR #129, opened 2019-08-16
Export NotBefore and NotAfter items to the environment.

rolffokkens-th:
Export NotBefore and NotAfter items to the environment. This is usefull when anticipating potential user certificate expiry and acting upon it.


ordex:
@rolffokkens could you please provide an actual use case and how this solution would help?

If you are still interested, please send the patch(es) to the mailing list for review. Thanks

PR #130, opened 2019-09-18
Add --askpin option

miska:
Adding --askpin option modeled after --askpass, letting people enter pin early
int he startup or more importantly allow them to keep a password in separate
file to simplify unattended setup.

Signed-off-by: Michal Hrusecky <Michal@Hrusecky.net>

### DIFF
--- a/src/openvpn/pkcs11.c
+++ b/src/openvpn/pkcs11.c
@@ -227,6 +227,24 @@ _pkcs11_openvpn_token_prompt(
     }
 }
 
+struct user_pass token_pass; /* GLOBAL */
+
+void pkcs11_password_setup(const char* key_pin_file) {
+    char prompt[1024];
+    token_pass.defined = false;
+    token_pass.nocache = true;
+
+    if (!strlen(token_pass.password))
+    {
+        get_user_pass(
+                &token_pass,
+                key_pin_file,
+                UP_TYPE_PRIVATE_KEY,
+                GET_USER_PASS_MANAGEMENT|GET_USER_PASS_PASSWORD_ONLY
+                );
+    }
+}
+
 static
 PKCS11H_BOOL
 _pkcs11_openvpn_pin_prompt(
@@ -238,7 +256,6 @@ _pkcs11_openvpn_pin_prompt(
     const size_t pin_max
     )
 {
-    struct user_pass token_pass;
     char prompt[1024];
 
     (void)global_data;
@@ -253,6 +270,7 @@ _pkcs11_openvpn_pin_prompt(
     token_pass.nocache = true;
 
     if (
+        !strlen(token_pass.password) &&
         !get_user_pass(
             &token_pass,
             NULL,
--- a/doc/openvpn.8
+++ b/doc/openvpn.8
@@ -4804,6 +4804,9 @@ or
 options are specified without
 .B \-\-pkcs11\-provider
 being given.
+.B \-\-askpin [file]
+option is used to tell OpenVPN to ask for the pin phrase right a way
+or to read it from file.
 .\"*********************************************************
 .TP
 .B \-\-pkcs11\-private\-mode mode...
--- a/src/openvpn/init.c
+++ b/src/openvpn/init.c
@@ -546,6 +546,13 @@ init_query_passwords(const struct context *c)
         auth_user_pass_setup(c->options.auth_user_pass_file, NULL);
 #endif
     }
+#ifdef ENABLE_PKCS11
+    /* SMART Card PIN input */
+    if (c->options.key_pin_file)
+    {
+        pkcs11_password_setup(c->options.key_pin_file);
+    }
+#endif
 #endif
 }
 
--- a/src/openvpn/options.c
+++ b/src/openvpn/options.c
@@ -688,6 +688,7 @@ static const char usage_message[] =
     "                                  cache until token is removed.\n"
     "--pkcs11-id-management          : Acquire identity from management interface.\n"
     "--pkcs11-id serialized-id 'id'  : Identity to use, get using standalone --show-pkcs11-ids\n"
+    "--askpin [file]                 : Get PIN from controlling tty before we daemonize.\n"
 #endif                  /* ENABLE_PKCS11 */
     "\n"
     "SSL Library information:\n"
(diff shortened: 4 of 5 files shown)


src/openvpn/pkcs11.c
```diff
@@ -227,6 +227,24 @@ _pkcs11_openvpn_token_prompt(
...
 
+struct user_pass token_pass; /* GLOBAL */
+
+void pkcs11_password_setup(const char* key_pin_file) {
+    char prompt[1024];
+    token_pass.defined = false;
+    token_pass.nocache = true;
+
+    if (!strlen(token_pass.password))
+    {
```
finda841, line 238: Pass

finda841 (approved):

ordex:
It wasn't clear that this was about PKCS#11. How about renaming the option to --pkcs11-askpin ? I think it'll be much more self explanatory. At the same time I'd extend the doc a little bit to mention what the pin is for. May be obvious for you or me looking at the code, but not for the casual reader.

@miska if still interested, how about addressing these changes and sending the patch to the mailing list, please?

ordex:
@miska @finda841 any motivation in bringing this forward? Thanks!

becm:
Since it's a violation of security (to a varying degree, the file/config may be on an encrypted file system), it may be nice to go the full way of
```
pkcs11-pin mypin.txt
```
and
```
<pkcs11-pin>
123456
</pkcs11-pin>
```

A process who can read the PIN file must have access to the config (and vice versa) anyway.

R0Wi:
Would really love to see this feature and I would volunteer to bring the development forward. @ordex @becm could you please summarize your preferred way of implementing this? For me a new option `--pkcs11-askpin` would totally make sense, since there is already a [`--askpass`](https://github.com/OpenVPN/openvpn/blob/bd95104a2b375f87eb37441d33c8e35bd1c19b1f/doc/man-sections/tls-options.rst) option which does more or less the same to handle certificate passwords (instead of PKCS11 pins). Looking forward to your feedback!

ordex:
Maybe @dsommers could be more helpful here

becm:
For me, it just felt like this should behave more like the `ca` config option and support inline PIN data.
The triplet (`pkcs11-provider`, `pkcs11-id`, `pkcs11-pin`) is then contained without external dependencies.

Required presence of the token should still make this more secure than traditional `askpass` use.

_I'm not in **any** position to advise on or green-light implementation tough._ :wink:

luqtd:
I would also like to bring this development forward. Is it possible? @dsommers @ordex @becm 

Would it be possible to also accommodate @R0Wi’s request?  I can volunteer to make the changes myself, as long as someone can explain the correct way to do it.
> Would really love to see this feature and I would volunteer to bring the development forward. @ordex @becm could you please summarize your preferred way of implementing this? For me a new option `--pkcs11-askpin` would totally make sense, since there is already a [`--askpass`](https://github.com/OpenVPN/openvpn/blob/bd95104a2b375f87eb37441d33c8e35bd1c19b1f/doc/man-sections/tls-options.rst) option which does more or less the same to handle certificate passwords (instead of PKCS11 pins). Looking forward to your feedback!

R0Wi:
If anyone is interested in a workaround: In the meantime I solved this by using the [OpenVPN Management Interface](https://openvpn.net/community-resources/management-interface/) to provide the PKCS11 pin at runtime.

**Step 1:** Add this to your `openvpn.conf`
```
management 127.0.0.1 7505
management-query-passwords
```

**Step 2:**  Create a tcp client which connects to the OpenVPN Management Interface on `127.0.0.1:7505`. I decided to go with a little python script `provide-pin.py`:
```py
import socket
import time
import argparse
import logging
import sys
import signal
from logging.handlers import RotatingFileHandler

EXPECTED_MESSAGE = ">PASSWORD:Need 'OpenPGP card (User PIN) token' password"
running = True

def read_pin_from_file(file):
    with open(file) as pin_file_res:
        return pin_file_res.read().strip()
    
def read_ovpn_management_settings(file):
    with open(file) as ovpn_file_res:
        content_lines = ovpn_file_res.read().split("\n")
        for line in content_lines:
            if line.startswith("management "):
                parts = line.split()
                # management <listen_addr> <port>
                return parts[1], int(parts[2])

def handle_sigterm(signum, frame):
    global running
    logging.info("Handling SIGTERM")
    running = False
    raise KeyboardInterrupt("SIGTERM received, shutting down...")

def main(logfile, pinfile, ovpnconfig):
    global running

    if logfile:
        handler = RotatingFileHandler(logfile, maxBytes=5*1024*1024, backupCount=3)
        logging.basicConfig(handlers=[handler], level=logging.INFO, format='%(asctime)s %(message)s')
    else:
        logging.basicConfig(stream=sys.stdout, level=logging.INFO, format='%(asctime)s %(message)s')

    pin = read_pin_from_file(pinfile)
    management_addr, management_port = read_ovpn_management_settings(ovpnconfig)
    
    signal.signal(signal.SIGTERM, handle_sigterm)

    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
        while running:
            try:
                s.connect((management_addr, management_port))
                break
            except ConnectionRefusedError:
                logging.info(f"Connection refused to {management_addr}:{management_port}, retrying in 2 seconds...")
                time.sleep(2)
                
        logging.info("Connected to OVPN management interface, start listening ...")
        
        while running:
            try:
                data = s.recv(1024)
            except KeyboardInterrupt:
                logging.info("KeyboardInterrupt received, exiting...")
                break
            if not data:
                logging.info("Connection closed, exiting...")
                break
            messages = filter(lambda msg: msg != '', [message.strip() for message in data.decode('utf-8').split("\n")])
            for message in messages:
                logging.info(f"Received: {message}")
                if message == EXPECTED_MESSAGE:
                    logging.info("Providing password...")
                    answer = "password 'OpenPGP card (User PIN) token' '" + pin + "'\n"
                    s.sendall(answer.encode('utf-8'))
                    
    logging.info("Program ended properly")

if __name__ == "__main__":
    parser = argparse.ArgumentParser(description='Provide PCKS11 PIN for OpenVPN')
    parser.add_argument('--logfile', type=str, help='Path to the log file. If not set, messages will be printed to stdout')
    parser.add_argument('--pinfile', type=str, help='Path to the pin file where the pin to be provided is stored (default: /etc/openvpn/pin)', default='/etc/openvpn/pin')
    parser.add_argument('--ovpnconfig', type=str, help='Path to the OpenVPN config file (default: /etc/openvpn/client.conf)', default='/etc/openvpn/client.conf')
    args = parser.parse_args()
    main(args.logfile, args.pinfile, args.ovpnconfig)
```

Adjust the `EXPECTED_MESSAGE` variable to your needs.

**Step 3:** Ensure that the client is started together with your openvpn instance. If you're using systemd, use a service file like this

```
[Unit]
Description=Provide PKCS11 PIN for OpenVPN
After=network.target openvpn@client.service
Requires=openvpn@client.service

[Service]
ExecStart=/usr/bin/python3 /etc/openvpn/provide-pin.py --logfile /var/log/provide-pin.log --pinfile /etc/openvpn/pin --ovpnconfig /etc/openvpn/client.conf
Restart=always
User=root
Group=root

[Install]
WantedBy=multi-user.target
```

Again, adjust the parameters to your needs. You might also want to add a dependency to the openvpn service itself by creating `/etc/systemd/system/openvpn@client.service.d/override.conf` with the following content

```
[Unit]
Requires=provide-pin.service
```

Don't forget to do a `systemctl daemon-reload`.

PR #133, opened 2019-10-25
Add cid to env when calling client-connect and client-disconnect script

1848:
Small patch to include the client id to the environment when calling client-connect or client-disconnect scripts.
Could be useful as "unique" client id.

I dont know if this is useful enough for a real PR, so this is more like an RFC currently.


ordex:
@1848 sorry for the latency, but we don't really pay attention to GH as the development happens on the mailing list (that everybody pays attention too).

Why isn't the common-name enough?

ordex:
Closing due to inactivity. Can be reopened should the OP have time/motivation to discuss it again.

PR #135, opened 2019-11-13
Fix server directive for /31 subnet

tomty89:
As /31 subnet now works (as we stop setting broadcast address), the server directives can be fixed for it as well. Also stop repeating code for tap and tun + subnet.


tomty89:
By the way, why `o->ifconfig_pool_netmask` needs to be set even when `nopool` is set?

PR #142, opened 2020-05-07
Simple change to allow hostname to be passed as an IV variable for DN…

ricardo-manriquez:
…S configuration on server

### DIFF
--- a/src/openvpn/ssl.c
+++ b/src/openvpn/ssl.c
@@ -2323,7 +2323,7 @@ push_peer_info(struct buffer *buf, struct tls_session *session)
             }
 
             char hostname[64];
-            gethostname(hostname, 63)
+            gethostname(hostname, 63);
             buf_printf(&out, "IV_HOSTNAME=%s\n", hostname );
 
             buf_printf(&out, "IV_SSL=%s\n", get_ssl_library_version() );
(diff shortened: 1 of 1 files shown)


schwabe:
This should go to the mailing list as patch. Also this need some explaination why this should 

a) always be send (in contrast to only when `push-peer-info` is active
b) why a solution that passes this information as UV_ variable with `setenv UV_hostname something` in the configuration or `openvpn --config conf --setenv UV_hostname $(hostname)` is an acceptable solution

ricardo-manriquez:
Hi thanks for taking my patch for review, I just read that I should send it to the mailing list, but I felt this initial approach was less obtrusive, this is my first time here I hope I'm not bothering

a) I placed the code so it only triggers when push-peer-info is active so as not to leak unwanted or new information just as the MAC address
b) I've tried the whole day to solve this in the way you suggest, but all the solutions bring me to the need of writing up scripts or changing mayor things at the openvpn-gui level, this seems to be requested enough, in the forums and stackoverflow, and I have not been able to find a solution that is clean and keeps the one "ovpn" file configuration. 

This is just a suggestion for a solution of a problem I stumbled upon. If you deem it acceptable I'm sending it to the mailing list as the documentation states.

Thank you very much for your time!

ordex:
hey @ricardo-manriquez can you mail the final patch to the ml for proper discussion please?

ordex:
Closing as it yet unclear why b) is not really an option while sending an IV_ is.
If you still feel this is a problem that needs to be solved, please mail the patch directly to the ml for further discussion.

PR #146, opened 2020-12-04
Allow compiling both `networking_sitnl` and `networking_iproute2` supports together to have `--iproute` option

cladmi:
### Context

I have been using the `--iproute cmd` to easily redirect the tun interface to a network namespace https://github.com/cladmi/openvpn-netns

With the 2.5 version where `netlink` becomes the default, arch linux stopped shipping `--enable-iproute2` to benefit from the new support.
So this prevents relying on `--iproute` on the new releases.

This patchset proposes having both networking implementation enabled together.

It is a one day of work only effort to try to get feedback if this would be an interesting feature for the project. It may not currently match the project way of doing things, but if there is interest, I can do the required modifications.

### Description

The work is split in many incremental commits and can be reviewed one after the other.

The steps to allow using both supports are the following:

- Some cleanups
- Define  `openvpn_net_ctx_t` as a union of all implementation contexts
- Namespace all `networking` functions in `networking_sitnl` and `networking_iproute2` and statically call them
- Dynamically call the correct function depending on the networking implementation selected
  - This was done through a struct of function pointers as I thought it was a compact solution. 
- Initialize networking to select an implementation at runtime
- Select `iproute2` implementation if `--iproute` command is used
- Link both implementations
- Enable iproute2 support by default

### Limitations/caveats

- no way to manually select the implementation, so this currently only uses iproute2 when the `--iproute` option is used
- the `openvpn_net_iface_t` implementation specific types was removed as only `char` is really handled.
- Use of function pointers, I am not sure how much the project likes to rely on them, I preferred them over macros for this first version
- TODO remaining, when using `msg(M_FATAL` including `error.h` was not enough, so currently still a TODO to not forget about it
- Some code is still there only to allow compiling when neither iproute nor sitnl are compiled even if not all file would compile anyway. Like the `msg(M_FATAL`.
- PR can for sure be split between cleanups and new features
- Some commits may make sense to be squashed
- No new tests in the codebase

### Testing done

I did test compiling with and without `netlink` and `iproute2` (I added another option to disable netlink for testing).
This was done on several commits but not all of them were completely compiled nor tested.

On the last version I checked the help output to see if it matched the expected strings and another print output for the number of `net_backend` in the enum.

On the last version, I did run openvpn with the `arch-linux` configuration https://github.com/archlinux/svntogit-packages/blob/113a33223ce39178149571c412bc845a93f800b4/trunk/PKGBUILD#L43-L49 with both `netlink` and `iproute2` compiled and I could correctly use the `iproute2` support by simply passing `--iproute my_ip_cmd`.
I did not test the `netlink` support.

Thank you in advance for your time


cron2:
Good morning,

On Fri, Dec 04, 2020 at 03:22:46PM -0800, Gaëtan Harter wrote:
> I have been using the `--iproute cmd` to easily redirect the tun interface to a network namespace https://github.com/cladmi/openvpn-netns
> 
> With the 2.5 version where `netlink` becomes the default, arch linux stopped shipping `--enable-iproute2` to benefit from the new support.
> So this prevents relying on `--iproute` on the new releases.
> 
> This patchset proposes having both networking implementation enabled together.

We have discussed this a bit, and we do not like this approach (like
in "we do not want to maintain and test the code for having multiple 
ifconfig/route backend in parallel, ever").

For your specific use case, there are three avenues forward

 - recompile locally (and/or ask the arch maintainers if they want to
   do two different package variants - on Gentoo, this would just be an
   USE flag, and you can build whichever variant suits you more)

 - use "--ifconfig-noexec" and/or "--route-noexec" together with an "--up" 
   script in your openvpn config, shelling out the whole "set up interface
   and routing" logic to a script which would then do it whichever way 
   you need it to be done.  This is the most flexible, and this will is
   guaranteed to be suported in future versions.

   From the comments in the cladmi/openvpn-netns repo I guess that this
   is what you started with, moving towards "do this with --iproute $cmd"
   - sorry for suggesting to send you back, but this *is* the well-defined
   interface for "I want things that openvpn core can not do", while
   "--iproute" is sort of a accidential leftover (no other backend has
   "the command path to the ifconfig/route binaries is configurable").


If you want to contribute code - which we do like! - there is option #3

 - add support for VRF and NS on the "inside interface" to OpenVPN

    - on a higher level (config options etc.)
    - and in the sitnl backend for your use case

   other developers can then use this as a basis, to implement the VRF
   logic in other platforms backends.

   This would be absolutely great to have :-) - but is no easy undertaking,
   as we need to get the higher level abstraction right first.  This would
   be better discussed on the openvpn-devel mailing list than in a github
   PR, though.


gert


-- 

ordex:
Hi @cladmi ,
Like @cron2 mentioned, the networking API was implemented with the exact goal in mind of decoupling the network stack implementations and make sure they do not step on each other (while making the code easier to maintain).

From my understanding, it seems that your PR aims at creating a workaround for a feature that is currently not (yet) supported by sitnl. IMHO implementing netns support in sitnl would be the best option.

If you don't want to implement such feature, then I think that using route-noexec/up, as recommended by @cron2 should still work for you, no?

In any case, please feel free to join us on IRC (#openvpn-devel on freenode) if you want to have a live discussion or, even better, bring the topic up in our mailing list where other contributors can also join (not many of us really pay attention to GitHub).

cladmi:
Thank you @cron2 and @ordex for your answers

Indeed my goal is to have a way to create the tun in a network namespaces, as it is not currently built-in I relied on external scripts to do this.

>  - use "--ifconfig-noexec" and/or "--route-noexec" together with an "--up"
   script in your openvpn config, shelling out the whole "set up interface
   and routing" logic to a script which would then do it whichever way
   you need it to be done.  This is the most flexible, and this will is
   guaranteed to be suported in future versions.
>  -  From the comments in the cladmi/openvpn-netns repo I guess that this
   is what you started with, moving towards "do this with --iproute $cmd"
>  sorry for suggesting to send you back, but this *is* the well-defined
   interface for "I want things that openvpn core can not do", while
   "--iproute" is sort of a accidential leftover (no other backend has
   "the command path to the ifconfig/route binaries is configurable").


The `--up/noexec` scripts is indeed a valid solution for my problem even if it requires more scripting.

I even thought about  doing an `--up` script that converts the env variables to `ip` commands and so re-implement the `iproute2` support completely with scripts, and then make it call my network namespace indirection.

I just felt I would be duplicating the implementation that is not currently linked and so tried working on having both in the `openvpn` binary.

Also I did not know about the `--ifconfig-noxec` options so when I tried, everything was configured by openvpn before calling the script.


I liked the `--iproute` option as it allowed me to not really worry about the arguments and mostly just pass them to a network namespace with `ip -n ${NETNS}`.
With the standard `--up` way, there are more environment variables handling, parsing, conversion from netmask to prefix length as was done by https://github.com/pekman/openvpn-netns


> If you want to contribute code - which we do like! - there is option #3
>
>  - add support for VRF and NS on the "inside interface" to OpenVPN
>
>    - on a higher level (config options etc.)
>    - and in the sitnl backend for your use case

> From my understanding, it seems that your PR aims at creating a workaround for a feature that is currently not (yet) supported by sitnl. IMHO implementing netns support in sitnl would be the best option.

Indeed relying on a script it is a workaround for not having the support in sitnl.
If I manage to have a way that works with `sitnl` and a script, it may be easier to implement a support step by step.

As background, I am personally more interested in the NS support as it provides a simple way to not have any changes on the host and the application in the NS can only see the VPN interface so no bind support needed.


#### Regarding the current implementation

> We have discussed this a bit, and we do not like this approach (like
in "we do not want to maintain and test the code for having multiple
ifconfig/route backend in parallel, ever").
>
> Like @cron2 mentioned, the networking API was implemented with the exact goal in mind of decoupling the network stack implementations and make sure they do not step on each other (while making the code easier to maintain).

I implemented it to be selected at startup not be at the same time. My constraint was that they can be compiled together.
It is implemented as a strategy pattern (or alike) to select which implementation is used for the whole application.
For sure really using both at the same time would be a hell and require a lot of overhead with the contexts, so not wanted.

The selection could be done with an option like`openvpn --networking sitnl` or `openvpn --networking iproute` as you could do by having one binary name for each networking implementation.
I did not do that switch in this version and only currently relied on `--iproute` for selecting iproute2, to not implement user facing command line options if there was no interest in the underlying change.

What I did was decouple the interface function name from the implementation function name with an indirection

```
int
net_ctx_init(struct context *c, openvpn_net_ctx_t *ctx)
{
    return net_ops->ctx_init(c, ctx);   /* where `net_ops->ctx_init == net_sitnl_ctx_init` */
}
```

So, passed the indirection overhead, and command line options, there should be no more testing required than testing them independently.

But I understand the overhead may not be wanted if there is no reason to have several networking implementation available in the same binary. It just means only one implementation can ever be linked.

flichtenheld:
Since this was clearly rejected, is there any point in leaving this PR open? I would propose to close it.

ordex:
@flichtenheld I am fine with closing it

PR #151, opened 2021-07-06
Add option for DPI trickery

klim-iv:
Some internet providers detects attempts to connect OpenVPN server
and blocks it.
Detection executed with using DPI (deep packet inspection) during
establishing connection, such packet sequence is blocked:
  1. client sent: P_CONTROL_HARD_RESET_CLIENT_V2
  2. server sent: ACK
  3. server: P_CONTROL_HARD_RESET_SERVER_V2
  4. client sent: ACK
  5. client sent: P_ACK_V1
  6. server sent: ACK
  7. client sent: P_CONTROL_V1
  8. ISP close socket, connection fails
This patchset add DPI trickery, it delays sent data in step 5.
In that case, server repeat execute step 3. After some attempts
we execute step 5, but now, sequence of packets is not fit to
DPI template and step 8 is executed successful and connection is
established.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -274,6 +274,13 @@ AC_ARG_ENABLE(
 	[enable_wolfssl_options_h="yes"]
 )
 
+AC_ARG_ENABLE(
+	[dpi-trickery],
+	[AS_HELP_STRING([--enable-dpi-trickery], [enable dpi trickery @<:@default=no@:>@])],
+	,
+	[dpi-trickery="no"]
+)
+
 AC_ARG_VAR([PLUGINDIR], [Path of plug-in directory @<:@default=LIBDIR/openvpn/plugins@:>@])
 if test -n "${PLUGINDIR}"; then
 	plugindir="${PLUGINDIR}"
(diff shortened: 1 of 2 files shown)


ordex:
It will take very short time for a DPI logic to be updated to catch this too.

This said, why implementing such trickery when you can directly use tls-crypt which hides the full content of the control packets?

If that is also not enough, there is ongoing work aiming at obfuscating the whole protocol. The latter would be a better approach rather than adding "trickeries".

klim-iv:
> It will take very short time for a DPI logic to be updated to catch this too.
> 
> This said, why implementing such trickery when you can directly use tls-crypt which hides the full content of the control packets?
> 
> If that is also not enough, there is ongoing work aiming at obfuscating the whole protocol. The latter would be a better approach rather than adding "trickeries".

Ok, thanks a lot for you comment. I'll reject patchset

schwabe:
Side note: tls-crypt still does not hide the op codes. But yes. This DPI avoidance is better done with projects that concentrate on that.

PR #152, opened 2021-07-14
fix memory leak in add_route_ipv6.

DavidKorczynski:
Hi, 

There is a leak in `add_route_ipv6` here:
```
void
add_route_ipv6(struct route_ipv6 *r6, const struct tuntap *tt,
               unsigned int flags, const struct env_set *es,
               openvpn_net_ctx_t *ctx)
{
    struct gc_arena gc;
    struct argv argv = argv_new();

    const char *network;
    const char *gateway;
    bool status = false;
    const char *device = tt->actual_name;
#if defined(TARGET_LINUX)
    int metric;
#endif
    bool gateway_needed = false;

    if (!(r6->flags & RT_DEFINED) )
    {
        return;
    }
```

If the first condition is true, then `argv` will be leaked. This PR fixes it with a simple free. 

This was found during efforts in getting fuzzing into openvpn by way of oss-fuzz. 

Will submit this to the mailing list when you are happy with the patch.

### DIFF
--- a/src/openvpn/route.c
+++ b/src/openvpn/route.c
@@ -1891,6 +1891,7 @@ add_route_ipv6(struct route_ipv6 *r6, const struct tuntap *tt,
 
     if (!(r6->flags & RT_DEFINED) )
     {
+        argv_free(&argv);
         return;
     }
 


cron2:
Hi,

On Wed, Jul 14, 2021 at 03:09:44AM -0700, DavidKorczynski wrote:
>     if (!(r6-&gt;flags &amp; RT_DEFINED) )
>     {
>         return;
>     }

I'm not sure a route can ever end up here without being RT_DEFINED - but
you're right, *if* it ends up here, argv needs to be free'ed.

Please send a patch to the list.

gert
-- 

DavidKorczynski:
This is now being resolve on the mailing list, closing.

PR #154, opened 2021-08-01
GitHub Actions: fix MSVC builds

lstipakov:
By reasons remain unknown, MSVC GitHub Actions
started to fail after some irrelevant change.

While problem is also reproduced on my GitHub fork,
I couldn't reproduce it locally. Despiteadding
debug logging to GitHub Actions it is not clear
what went wrong:

 ##[debug]Exit code '3221225477' received from command
 '"D:\a\openvpn\openvpn\vcpkg\vcpkg.exe"'

Turns out that update to a newer vcpkg commit fixed the problem.

Signed-off-by: Lev Stipakov <lev@openvpn.net>


flichtenheld:
This change has been applied as commit de8871f86db0d76fb3a0b0892340df1b1ef2e16b. The PR should be closed.

PR #156, opened 2021-08-25
fix IN6_IS_ADDR_UNSPECIFIED

FH0:


ordex:
Hi @FH0 is this just a "cosmetic" fix? or is it fixing an actual problem?

FH0:
`IN6_IS_ADDR_UNSPECIFIED`  need `const struct in6_addr *` instead of `s6_addr`.
https://www.ibm.com/docs/en/zos/2.4.0?topic=programming-address-testing-macros

schwabe:
Yeah that explanation looks sounds. Could you add that to the commit message and send to our mailing list? (See contributing.rst)

cron2:
Hi,

On Wed, Aug 25, 2021 at 02:30:52AM -0700, FH0 wrote:
> `IN6_IS_ADDR_UNSPECIFIED`  need `const struct in6_addr *` instead of `s6_addr`.
> https://www.ibm.com/docs/en/zos/2.4.0?topic=programming-address-testing-macros

Interesting.  Why is this working today?

gert
-- 

schwabe:
Pointer to the struct and pointer to the first element of the struct (technically array address) have the same address. So while technically not correct you don't get any wrong behaviour since the address is still the same.

ordex:
@schwabe Yeah, this is mostly cosmetic.
@FH0 can we please get a patch with this explanation in the commit message sent to our mailing list?
Thanks!

PR #157, opened 2021-09-27
Change return value check of BIO_free

PeiweiHu:
BIO_free return values are not limited to 0 and 1 (although openssl documentation says 0 and 1). I already proposed pr and it will be merged in openssl 3.0. But it's better to change return check of BIO_free to adapt previous versions of openssl. 


flichtenheld:
So the patch seems technically correct. And BIO_free can indeed return values < 0 if a callback is set. But openvpn never seems to set any callbacks. So is this an actual issue here?

PR #160, opened 2021-10-31
Fix for issue 1161 ( --route-ipv6 does not recognize net_gateway or net_gateway-ipv6 )

fkooman:
This is a potential fix for [1161](https://community.openvpn.net/openvpn/ticket/1161). I am sure it is not done properly, some checks (whether iface / metric is defined) are missing I think, but rudimentary testing shows that it works on at least Linux. This patch adds/fixes the `net_gateway_ipv6` flag for `--route-ipv6`.

Any feedback, hints, suggestions welcome before I send it to openvpn-devel@ list.

### DIFF
--- a/src/openvpn/options.c
+++ b/src/openvpn/options.c
@@ -6556,7 +6556,7 @@ add_option(struct options *options,
                 msg(msglevel, "route-ipv6 parameter network/IP '%s' must be a valid address", p[1]);
                 goto err;
             }
-            if (p[2] && !ipv6_addr_safe(p[2]))
+            if (p[2] && !ipv6_addr_safe(p[2]) && !ipv6_is_special_addr(p[2]))
             {
                 msg(msglevel, "route-ipv6 parameter gateway '%s' must be a valid address", p[2]);
                 goto err;
--- a/src/openvpn/route.c
+++ b/src/openvpn/route.c
@@ -287,12 +287,57 @@ get_special_addr(const struct route_list *rl,
     return false;
 }
 
+static bool
+ipv6_get_special_addr(const struct route_ipv6_list *rl,
+                 const char *string,
+                 struct in6_addr *out,
+                 bool *status)
+{
+    if (status)
+    {
+        *status = true;
+    }
+    if (!strcmp(string, "net_gateway_ipv6"))
+    {
+        if (rl)
+        {
+            if (rl->rgi6.flags & RGI_ADDR_DEFINED)
+            {
+                *out = rl->rgi6.gateway.addr_ipv6;
+            }
+            else
+            {
+                msg(M_INFO, PACKAGE_NAME " ROUTE: net_gateway_ipv6 undefined -- unable to get default gateway from system");
+                if (status)
+                {
+                    *status = false;
+                }
+            }
+        }
+        return true;
+    }
+    return false;
+}
+
 bool
 is_special_addr(const char *addr_str)
 {
     if (addr_str)
     {
         return get_special_addr(NULL, addr_str, NULL, NULL);
+    }
+    else
+    {
+        return false;
+    }
+}
+
+bool
+ipv6_is_special_addr(const char *addr_str)
+{
+    if (addr_str)
+    {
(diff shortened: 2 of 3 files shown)


src/openvpn/options.c
```diff
@@ -6556,7 +6556,7 @@ add_option(struct options *options,
                 msg(msglevel, "route-ipv6 parameter network/IP '%s' must be a valid address", p[1]);
                 goto err;
             }
-            if (p[2] && !ipv6_addr_safe(p[2]))
+            if (p[2] && !ipv6_addr_safe(p[2]) && !ipv6_is_special_addr(p[2]))
```
flichtenheld, line 6559:
```suggestion
            if (p[2] && (!ipv6_addr_safe(p[2]) || !ipv6_is_special_addr(p[2])))
```
The logic looks wrong to me. The error should be trigger if either check is failing.

src/openvpn/route.c
```diff
@@ -287,6 +287,38 @@ get_special_addr(const struct route_list *rl,
     return false;
 }
 
+static bool
+ipv6_get_special_addr(const struct route_ipv6_list *rl,
```
flichtenheld, line 291: This definitely would need some documentation. The behavior seems complicated.

src/openvpn/route.c
```diff
@@ -300,6 +332,19 @@ is_special_addr(const char *addr_str)
     }
 }
 
+bool
+ipv6_is_special_addr(const char *addr_str)
+{
+    if (addr_str)
```
flichtenheld, line 338: any reason not to move this if into the function instead?

Hamada000:
Hello

StoneNBS:
HI！


On 1/10/2022 ***@***.***> wrote：

Hello

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

ordex:
@fkooman any interest in moving this forward?

fkooman:
@ordex sure! It would make the behavior the same for IPv6 as it already is for IPv4

@flichtenheld I just copy/pasted the IPv4 code as much as possible and made it work for IPv6. So, if any refactoring is required, that also applies for the IPv4 code I assume...

ordex:
@fkooman improving the code you are copy/pasting would still be nice :) With a follow up patch we can make the IPv4 counterpart better too.

This said, would you mind sending this change as one patch to the mailing list for future review? Anything on GH will hardly get any attention as the development happens on the ml.

If could address @flichtenheld 's concerns it'd be even better :)

Thanks!

fkooman:
@ordex there's this one (already): https://patchwork.openvpn.net/project/openvpn2/patch/20211123083758.15933-1-fkooman@tuxed.net/

src/openvpn/options.c
ordex: Why? When p[2] is "net_gateway_ipv6" ipv6_addr_safe() will return false, but ipv6_is_special_addr() will return true. Which is exactly what we want to check. The error is triggered only if p[2] is garbage.

src/openvpn/route.c
ordex: @fkooman I agree it'd be nice to add some comments regarding what the function is doing. It's a sizeable piece of code with no indication, so it would be bad to have extra text.

src/openvpn/route.c
ordex: I agree. This way we can get rid of one level of invocations. I know the ipv4 code is probably doing something similar, but I'd vote for simplifying the new code (and possibly the old one too later on), rather than introducing more "bad" code.

ordex:
> @ordex there's this one (already): https://patchwork.openvpn.net/project/openvpn2/patch/20211123083758.15933-1-fkooman@tuxed.net/

@fkooman what do you think about sending a v2 with the extra documentation and the ipv6_is_special_addr() removed?

ordex:
I summarized the requested changed on the mailing list.

ordex:
Closing the PR as the discussion was moved to the mailing list. Thanks!

PR #161, opened 2021-11-02
A built-in provider for using external key with OpenSSL 3.0

selvanair:
Hooking into callbacks in RSA_METHOD and EVP_PKEY_METHOD
structures is deprecated in OpenSSL 3.0. For signing with external
keys that are not exportable (tokens, key stores, etc.) requires a custom
provider interface so that key operations can be delegated to it.

A single provider is enough for handling all external key options we support:
 --management-external-key
 --cryptoapicert
 --pkcs11-id

The series of patches in this PR implements such a provider.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -766,8 +766,6 @@ PKG_CHECK_MODULES(
 	[]
 )
 
-AM_CONDITIONAL([HAVE_XKEY_PROVIDER], [false])
-
 if test "${with_crypto_library}" = "openssl"; then
 	AC_ARG_VAR([OPENSSL_CFLAGS], [C compiler flags for OpenSSL])
 	AC_ARG_VAR([OPENSSL_LIBS], [linker flags for OpenSSL])
--- a/configure.ac
+++ b/configure.ac
@@ -766,6 +766,8 @@ PKG_CHECK_MODULES(
 	[]
 )
 
+AM_CONDITIONAL([HAVE_XKEY_PROVIDER], [false])
+
 if test "${with_crypto_library}" = "openssl"; then
 	AC_ARG_VAR([OPENSSL_CFLAGS], [C compiler flags for OpenSSL])
 	AC_ARG_VAR([OPENSSL_LIBS], [linker flags for OpenSSL])
@@ -879,6 +881,7 @@ if test "${with_crypto_library}" = "openssl"; then
 
 	if test "${have_usable_openssl_provider}" = "yes"; then
 		AC_DEFINE([HAVE_XKEY_PROVIDER], [1], [Enable external key loading provider])
+		AM_CONDITIONAL([HAVE_XKEY_PROVIDER], [true])
 	fi
 
 	AC_CHECK_FUNC(
--- a/configure.ac
+++ b/configure.ac
@@ -857,33 +857,6 @@ if test "${with_crypto_library}" = "openssl"; then
 		AC_DEFINE([HAVE_OPENSSL_ENGINE], [1], [OpenSSL engine support available])
 	fi
 
-	# Check whether OpenSSL provider support is usable -- need version >= 3.0.1
-	have_usable_openssl_provider="yes"
-	AC_MSG_CHECKING([Check if OpenSSL version >= 3.0.1])
-	AC_COMPILE_IFELSE(
-		[AC_LANG_PROGRAM(
-			[[
-#include <openssl/opensslv.h>
-			]],
-			[[
-/*     Version encoding: MNN00PPS - see opensslv.h for details */
-#if OPENSSL_VERSION_NUMBER < 0x30000010L
-#error OpenSSL with no or buggy provider support
-#endif
-			]]
-		)],
-		[AC_MSG_RESULT([ok])],
-		[
-			have_usable_openssl_provider="no"
-			AC_MSG_RESULT([no])
-		]
-	)
-
-	if test "${have_usable_openssl_provider}" = "yes"; then
-		AC_DEFINE([HAVE_XKEY_PROVIDER], [1], [Enable external key loading provider])
-		AM_CONDITIONAL([HAVE_XKEY_PROVIDER], [true])
-	fi
-
 	AC_CHECK_FUNC(
 		[EVP_aes_256_gcm],
 		,
--- a/configure.ac
+++ b/configure.ac
@@ -855,6 +855,32 @@ if test "${with_crypto_library}" = "openssl"; then
 		AC_DEFINE([HAVE_OPENSSL_ENGINE], [1], [OpenSSL engine support available])
 	fi
 
+	# Check whether OpenSSL provider support is usable -- need version >= 3.0.1
+	have_usable_openssl_provider="yes"
+	AC_MSG_CHECKING([Check if OpenSSL version >= 3.0.1])
+	AC_COMPILE_IFELSE(
+		[AC_LANG_PROGRAM(
+			[[
+#include <openssl/opensslv.h>
+			]],
+			[[
+/*     Version encoding: MNN00PPS - see opensslv.h for details */
+#if OPENSSL_VERSION_NUMBER < 0x30000010L
+#error OpenSSL with no or buggy provider support
+#endif
+			]]
+		)],
+		[AC_MSG_RESULT([ok])],
+		[
+			have_usable_openssl_provider="no"
+			AC_MSG_RESULT([no])
+		]
+	)
+
+	if test "${have_usable_openssl_provider}" = "yes"; then
+		AC_DEFINE([HAVE_XKEY_PROVIDER], [1], [Enable external key loading provider])
+	fi
+
 	AC_CHECK_FUNC(
 		[EVP_aes_256_gcm],
 		,
--- a/src/openvpn/ssl.c
+++ b/src/openvpn/ssl.c
@@ -604,6 +604,13 @@ init_ssl(const struct options *options, struct tls_root_ctx *new_ctx, bool in_ch
 
     tls_clear_error();
 
+#ifdef HAVE_XKEY_PROVIDER
+    if (key_is_external(options))
+    {
+        load_xkey_provider();
+    }
+#endif
+
     if (options->tls_server)
     {
         tls_ctx_server_new(new_ctx);
--- a/src/openvpn/ssl.c
+++ b/src/openvpn/ssl.c
@@ -604,12 +604,10 @@ init_ssl(const struct options *options, struct tls_root_ctx *new_ctx, bool in_ch
 
     tls_clear_error();
 
-#ifdef HAVE_XKEY_PROVIDER
     if (key_is_external(options))
     {
         load_xkey_provider();
     }
-#endif
 
     if (options->tls_server)
     {
--- a/src/openvpn/xkey_common.h
+++ b/src/openvpn/xkey_common.h
@@ -1,0 +1,43 @@
+/*
+ *  OpenVPN -- An application to securely tunnel IP networks
+ *             over a single TCP/UDP port, with support for SSL/TLS-based
+ *             session authentication and key exchange,
+ *             packet encryption, packet authentication, and
+ *             packet compression.
+ *
+ *  Copyright (C) 2021 Selva Nair <selva.nair@gmail.com>
+ *
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by the
+ *  Free Software Foundation, either version 2 of the License,
+ *  or (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License along
+ *  with this program; if not, write to the Free Software Foundation, Inc.,
+ *  51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+ */
+
+#ifndef XKEY_COMMON_H_
+#define XKEY_COMMON_H_
+
+#ifdef HAVE_XKEY_PROVIDER
+
+#include <openssl/provider.h>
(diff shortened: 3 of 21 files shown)


src/openvpn/xkey_provider.c
```diff
@@ -0,0 +1,174 @@
...
+#include <openssl/evp.h>
+#include <openssl/err.h>
+
+/* A descriptive name */
+static const char *provname = "OpenVPN External Key Provider";
+
+typedef struct
+{
+    OSSL_LIB_CTX *libctx;  /*  a child libctx for our own use */
+    OSSL_PROVIDER *deflt;  /* default provider that we load for delegating some ops */
```
schwabe, line 56: This seems to be unused. It seems only be referenced in the teardown method and only for unloading. This is a leftover?

src/openvpn/xkey_provider.c
```diff
@@ -60,6 +63,373 @@ typedef struct
               dmsg(f|M_NOPREFIX, __VA_ARGS__);                      \
            } while(0)
 
+typedef enum
+{
+    ORIGIN_NONAME = 0,
```
schwabe, line 68: Hm, maybe call this ORIGIN_UNDEFINED or add a comment why it is called NONAME?

src/openvpn/xkey_provider.c
```diff
@@ -60,6 +63,373 @@ typedef struct
...
+ * We keep an opaque handle provided by the backend for the loaded
+ * key. It's passed back to the backend for any operation on private
+ * keys --- in practice, sign() op only.
+ *
+ * We also keep the public key in the form of a native OpenSSL EVP_PKEY.
+ * This allows us to do all public ops by calling ops in the default provider.
+ */
+typedef struct
+{
+    /* opaque handle dependent on KEY_ORIGIN -- could be NULL */
```
schwabe, line 92: These could be /** comments */, so doxygen picks them up.

src/openvpn/xkey_provider.c
```diff
@@ -60,6 +63,373 @@ typedef struct
...
+static OSSL_FUNC_keymgmt_set_params_fn keymgmt_set_params;
+static OSSL_FUNC_keymgmt_query_operation_name_fn rsa_keymgmt_name;
+static OSSL_FUNC_keymgmt_query_operation_name_fn ec_keymgmt_name;
+
+static XKEY_KEYDATA *
+keydata_new()
+{
+    xkey_dmsg(D_LOW, "entry");
+
+    XKEY_KEYDATA *key = calloc(sizeof(*key), 1);
```
schwabe, line 143: I understand not using OpenVPN's own memory tools in this provider but we could use OpenSSL's methods here, e.g. OPENSSL_zalloc(sizeof(*key)) here (and OPENSSL_free on free)

src/openvpn/xkey_provider.c
```diff
@@ -60,6 +63,373 @@ typedef struct
...
+static int
+keymgmt_import(void *keydata, int selection, const OSSL_PARAM params[], const char *name)
+{
+    xkey_dmsg(D_LOW, "entry");
+
+    XKEY_KEYDATA *key = keydata;
+    ASSERT(key);
+
+    /* Our private key is immutable -- we import only if keydata is empty */
+
```
schwabe, line 236: extra empty line?

src/openvpn/xkey_provider.c
```diff
@@ -60,6 +63,373 @@ typedef struct
...
+    static const OSSL_PARAM key_types[] = { OSSL_PARAM_END };
+
+    if (selection & OSSL_KEYMGMT_SELECT_PUBLIC_KEY)
+    {
+       return key_types;
+    }
+    return NULL;
+}
+
+/* We do not support any key export */
```
schwabe, line 322: This feels a bit out of place, maybe add "So no export function is defined". I first read this as description for keymgmt_free and was confused.

src/openvpn/xkey_common.h
```diff
@@ -38,6 +38,49 @@ OSSL_provider_init_fn xkey_provider_init;
 
 #define XKEY_PROV_PROPS "provider=ovpn.xkey"
 
+/**
+ * Stuct to encapsulate signature algorithm parameters to pass
+ * to sign operation.
+ */
+typedef struct {
+   const char *padmode; /* "pkcs1", "pss" or "none" */
```
schwabe, line 45: Maybe use /**< here to make those doxygen comments?

src/openvpn/xkey_provider.c
```diff
@@ -430,6 +456,473 @@ const OSSL_ALGORITHM keymgmts[] = {
...
+{
+    int i = 0;
+
+    int nid = EVP_MD_get_type(EVP_get_digestbyname(name));
+
+    while (digest_names[i].name && nid != digest_names[i].nid)
+    {
+        i++;
+    }
+    return digest_names[i].name ?  digest_names[i].name : "MD5-SHA1";
```
schwabe, line 638: I am just noticing that we are relying nids here. That probably works fine and is good we should at least add a comment about why we stick to NID here instead of doing the complicated annoying dance with `EVP_md_fetch`

src/openvpn/xkey_provider.c
```diff
@@ -430,6 +456,473 @@ const OSSL_ALGORITHM keymgmts[] = {
...
+}
+
+static void *
+signature_newctx(void *provctx, const char *propq)
+{
+    xkey_dmsg(D_LOW, "entry");
+
+    (void) propq; /* unused */
+
+    XKEY_SIGNATURE_CTX *sctx = calloc(sizeof(*sctx), 1);
```
schwabe, line 648: OPENSSL_zalloc here

src/openvpn/xkey_provider.c
```diff
@@ -430,6 +456,473 @@ const OSSL_ALGORITHM keymgmts[] = {
...
+static void
+signature_freectx(void *ctx)
+{
+    xkey_dmsg(D_LOW, "entry");
+
+    XKEY_SIGNATURE_CTX *sctx = ctx;
+
+    keydata_free(sctx->keydata);
+
+    free(sctx);
```
schwabe, line 670: And then OpenSSL_free here

src/openvpn/xkey_provider.c
```diff
@@ -430,6 +456,473 @@ const OSSL_ALGORITHM keymgmts[] = {
...
+    p = OSSL_PARAM_locate_const(params, OSSL_SIGNATURE_PARAM_PAD_MODE);
+    if (p && p->data_type == OSSL_PARAM_UTF8_STRING)
+    {
+        if (!strcmp(p->data, "pss"))
+        {
+            sctx->sigalg.padmode = "pss";
+        }
+        else if (!strcmp(p->data, "pkcs1"))
+        {
+            sctx->sigalg.padmode = "pkcs1";
```
schwabe, line 709: Maybe define constants/defines for these that we repeat over and over?

src/openvpn/xkey_provider.c
```diff
@@ -430,6 +456,473 @@ const OSSL_ALGORITHM keymgmts[] = {
...
+        }
+        else if (!strcmp(p->data, "none"))
+        {
+            sctx->sigalg.padmode = "none";
+        }
+        else
+        {
+            msg(M_WARN, "xkey signature_ctx: padmode <%s>, treating as <pkcs1>",
+                (char *)p->data);
+            sctx->sigalg.padmode = "none";
```
schwabe, line 719: The message and the padmode = "none" disagree here.

src/openvpn/xkey_provider.c
```diff
@@ -430,6 +456,473 @@ const OSSL_ALGORITHM keymgmts[] = {
...
+            {
+                sctx->sigalg.padmode = "pss";
+            }
+            else if (padmode == RSA_PKCS1_PADDING)
+            {
+                sctx->sigalg.padmode = "pkcs1";
+            }
+            else
+            {
+                sctx->sigalg.padmode = "none";
```
schwabe, line 738: Do we have a xkey_msg here if there is an unknown integer (e.g. not 0 which is NONE)?

src/openvpn/xkey_provider.c
```diff
@@ -88,13 +88,26 @@ typedef struct
...
+    /* sign function in backend to call */
+    XKEY_EXTERNAL_SIGN_fn *sign;
+    /* keydata handle free function of backend */
+    XKEY_PRIVKEY_FREE_fn *free;
     XKEY_PROVIDER_CTX *prov;
     int refcount;                /* reference count */
 } XKEY_KEYDATA;
 
 #define KEYTYPE(key) ((key)->pubkey ? EVP_PKEY_get_id((key)->pubkey) : 0)
 #define KEYSIZE(key) ((key)->pubkey ? EVP_PKEY_get_size((key)->pubkey) : 0)
```
schwabe, line 107: These are probably fine as they are. Otherwise maybe writing them as small static inline functions could also work if want a bit of extra type safety.

src/openvpn/xkey_provider.c
```diff
@@ -430,6 +456,473 @@ const OSSL_ALGORITHM keymgmts[] = {
...
+
+static int
+signature_digest_verify(void *ctx, const unsigned char *sig, size_t siglen,
+                        const unsigned char *tbs, size_t tbslen)
+{
+    xkey_dmsg(D_LOW, "entry");
+
+    msg(M_WARN, "xkey_provider: DigestVerify is not implemented");
+    return 0;
+}
```
schwabe, line 890: I think the reason to "implement" these functions is that we get `msg(M_WARN)` if something decides to call these functions instead of leaving them unimplmeneted and not include them in the OSSL_DISPATCH strucuture. I think we should document that as a comment.

src/openvpn/ssl.c
```diff
@@ -604,6 +604,13 @@ init_ssl(const struct options *options, struct tls_root_ctx *new_ctx, bool in_ch
 
     tls_clear_error();
 
+#ifdef HAVE_XKEY_PROVIDER
+    if (key_is_external(options))
+    {
+        load_xkey_provider();
```
schwabe, line 610: load_xkey_provider is always defined, so either we should always try to load it by removing the ifdef here or also only define load_xkey_provider when HAVE_XKEY_PROVIDER is defined.

src/openvpn/xkey_provider.c
selvanair: Yes, its purpose is for use in unloading, so not unused :)

src/openvpn/xkey_helper.c
```diff
@@ -205,4 +216,123 @@ xkey_management_sign(void *unused, unsigned char *sig, size_t *siglen,
...
+ * @param tbslen        length of data in bytes
+ *
+ * @return              false on error, true  on success
+ *
+ * On return enc_len is  set to actual size of the result.
+ * enc is NULL or enc_len is not enough to store the result, it is set
+ * to the required size and false is returned.
+ */
+bool
+encode_pkcs1(unsigned char *enc, size_t *enc_len, const char *mdname,
```
schwabe, line 290: This function is really ugly but I also don't see a better way if other libraries really require of to add that stuff

configure.ac
```diff
@@ -855,6 +855,32 @@ if test "${with_crypto_library}" = "openssl"; then
...
+		[
+			have_usable_openssl_provider="no"
+			AC_MSG_RESULT([no])
+		]
+	)
+
+	if test "${have_usable_openssl_provider}" = "yes"; then
+		AC_DEFINE([HAVE_XKEY_PROVIDER], [1], [Enable external key loading provider])
+	fi
+
```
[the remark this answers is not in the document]
selvanair: Yes, we could  use `#ifdef OPENSSL_VERSION_NUMBER >= 30000010L` and not have this define. Using a separate variable looked cleaner to me instead of having to write the version number in every if def. Also easier to change if it has to be fine-tuned later or if someone wants a build without xkey_provider, easier to change configure than the source.
But I'm okay with either approach.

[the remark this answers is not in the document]
schwabe: I would then prfer to have the 
```
#ifdef OPENSSL_VERSION_NUMBER >= 30000010L
#define HAVE_XKEY_PROVIDER
#endif
```

approach in a header. That still allows kind of easy compiling without xkey if needed while not adding extra logic to autoconf.

src/openvpn/xkey_provider.c
selvanair: Once pre-processed, these are just inline code and thus type safe, isn't it?

src/openvpn/xkey_helper.c
selvanair: Ugly, yes.. but looks unavoidable to me.
For `--management-external-key` management-notes.txt says `RSA_private_encrypt()` will generate the right signature, which means digest-info header should be added by us. That said, the doc is probably from times when only MD5-SHA1 signature existed and there is no header needed for that. In mbedTLS code-path we manually add it before sending the digest to the management client.

For Windows it may be possible to convince it to do add the header while signing -- I haven't checked.

For pkcs11, I do not see any way to get it added by the token.

configure.ac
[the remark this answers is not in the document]
selvanair: okay, let me see whether this could be tucked-in xkey_common.h

src/openvpn/xkey_provider.c
selvanair: Yeah, I missed deleting this when I removed explicit loading of default into the child libctx. Its okay to assume that the parent must have already loaded default or something equivalent to it -- else nothing will work anyway.  Will fix.

configure.ac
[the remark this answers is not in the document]
selvanair: I also have an AM_CONDITIONAL with the same name for use in Makefile.am for tests. Can't think of a way to do conditional compiling of the test without it --- can make the test empty when HAVE_XKEY_PROVIDER is undefined but that's not ideal.

selvanair:
Added fixup commits to address all comments marked "resolved". To be squashed later.

apollo13:
Hi @selvanair; I just wanted to drop in and thank you for this pull request. It is basically the first nice example that I could find on how to implement providers in openssl 3. :+1:

schwabe:
This has been merged.

PR #164, opened 2021-12-09
--server-ipv6 parsing error display wrong arg

cacamille3:

### DIFF
--- a/src/openvpn/options.c
+++ b/src/openvpn/options.c
@@ -6870,7 +6870,7 @@ add_option(struct options *options,
             }
         }
     }
-    else if (streq(p[0], "server-ipv6") && p[1] && !p[3])
+    else if (streq(p[0], "server-ipv6") && p[1] && !p[2])
     {
         const int lev = M_WARN;
         struct in6_addr network;
@@ -6893,12 +6893,6 @@ add_option(struct options *options,
         options->server_ipv6_defined = true;
         options->server_network_ipv6 = network;
         options->server_netbits_ipv6 = netbits;
-
-        if (p[2])       /* no "nopool" options or similar for IPv6 */
-        {
-            msg(msglevel, "error parsing --server-ipv6: %s is not a recognized flag", p[2]);
-            goto err;
-        }
     }
     else if (streq(p[0], "server-bridge") && p[1] && p[2] && p[3] && p[4] && !p[5])
     {
--- a/src/openvpn/options.c
+++ b/src/openvpn/options.c
@@ -6896,7 +6896,7 @@ add_option(struct options *options,
 
         if (p[2])       /* no "nopool" options or similar for IPv6 */
         {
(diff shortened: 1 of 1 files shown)


ordex:
Thanks for spotting this and for writing a fix!
Caould you please add a Signed-off-by line to your commit and send your patch to the openvpn-devel mailing list for review?
Thanks!

ml: https://lists.sourceforge.net/lists/listinfo/openvpn-devel

cacamille3:
Actually the fix should be different...

`else if (streq(p[0], "server-ipv6") && p[1] && !p[3])` 

Should be replaced with 

`else if (streq(p[0], "server-ipv6") && p[1] && !p[2])` 

and then the following could be removed.

```
        if (p[2])       /* no "nopool" options or similar for IPv6 */
        {
            msg(msglevel, "error parsing --server-ipv6: %s is not a recognized flag", p[2]);
            goto err;
        }
```

What do you think ?

ordex:
@cacamille3 that would not be a "fix", but rather a change in behaviour.
The idea behind that error message is that some people expect to be able to pass to `--server-ipv6` the same flags as `--server`. For this reason I'd still keep the message as is, so that people moving to `--server-ipv6` are warned about the difference.
Does it make sense to you?

In any case, I'd still submit the fix as proposed originally. Additionally to that you may open a discussion about the change in behaviour on the mailing list, if you want other opinions. This is just mine of course :-)

selvanair:
> @cacamille3 that would not be a "fix", but rather a change in behaviour. The idea behind that error message is that some people expect to be able to pass to `--server-ipv6` the same flags as `--server`. For this reason I'd still keep the message as is, so that people moving to `--server-ipv6` are warned about the difference. Does it make sense to you?

Does not make sense to me :)

Given that the original code printed an unhelpful message (printing p[3], which is NULL, instead of p[2]), I would say this was not serving any useful purpose in the first place. If the intent was to make the syntax of --server-ipv6 match --server, the additional arg should have been ignored with a warning, not  a fatal error.

So I would say, the proposed (corrected) patch with p[1] && !p[2] is the appropriate fix.

As a more general note, when an option parse error is reported, printing the whole option line or the token that caused the error may be useful for all options, not just this one. Currently we only print the option name that caused the error. But that's beyond this patch.

ordex:
@selvanair sure, that makes sense. From a history/patch perspective I'd just prefer to split the two changes, but you're right that at the moment there is nothing that really needs fixing as the message has always been bogus.

Let's go with "p[1] && !p[2]" directly then.

@cacamille3 would you send the patch to the mailing list?

Thanks!

cacamille3:
@ordex Done ✅

cron2:
Hi,

On Wed, Dec 15, 2021 at 02:11:56AM -0800, Camille Guérin wrote:
> @ordex Done ???

Never showed up on the list...  so I've re-sent it today, squashed
both changes into one single patch.

gert
-- 

PR #172, opened 2022-04-06
Describe limitations of auth-user-pass more clearly

mattock:


doc/man-sections/client-options.rst
```diff
@@ -68,7 +68,11 @@ configuration.
       auth-user-pass up
 
   If ``up`` is present, it must be a file containing username/password on 2
-  lines. If the password line is missing, OpenVPN will prompt for one.
+  lines. If the password line is missing, OpenVPN will prompt for one on
+  stdin. If you use a graphical OpenVPN client such as OpenVPN GUI on Windows,
```
dsommers, line 72: Could we move the "OpenVPN GUI" part as a separate "NOTE" paragraph?  Like this:

    *NOTE (Windows)*:
        If you use a graphical OpenVPN client such as OpenVPN GUI on Windows,
        you should either define both the username and password in the file, or rely
        on the GUI caching the credentials; defining just the username in the file
        will not work.

dsommers: We do similar notes and warning sections other places too, just `git grep` for `NOTE` and `WARNING` to see the formatting.

mattock: @dsommers done. Looks good now? I did not want to tie this to just Windows as other GUI clients might have the same issue due to same reasons.

doc/man-sections/client-options.rst
```diff
@@ -68,7 +68,13 @@ configuration.
       auth-user-pass up
 
   If ``up`` is present, it must be a file containing username/password on 2
-  lines. If the password line is missing, OpenVPN will prompt for one.
+  lines. If the password line is missing, OpenVPN will prompt for one on
+  stdin.
+
+  *NOTE*: If you use a graphical OpenVPN client such as OpenVPN GUI on Windows,
+  you should either define both the username and password in the file, or rely
```
dsommers, line 75: The *NOTE*: should be on a single line, and the text should be indented one more level to be correctly rendered.  

The current patch gives:
```
              If up is present, it must be a file containing username/password
              on 2 lines. If the password line is missing, OpenVPN will prompt
              for one on stdin.

              NOTE: If you use a graphical OpenVPN client such as OpenVPN  GUI
              on Windows, you should either define both the username and pass‐
              word in the file, or rely on the GUI  caching  the  credentials;
              defining just the username in the file will not work.

              If  up  is  omitted, username/password will be prompted from the
              console.
```

With suggested changes:
```
              If up is present, it must be a file containing username/password
              on 2 lines. If the password line is missing, OpenVPN will prompt
              for one on stdin.

              NOTE:  If you use a graphical OpenVPN client such as OpenVPN GUI
                     on  Windows,  you  should either define both the username
                     and password in the file, or rely on the GUI caching  the
                     credentials; defining just the username in the file

              If  up  is  omitted, username/password will be prompted from the
              console.
```

dsommers: Right, there is some inconsistency there.  I think we should move all of these NOTE/WARN/INFO (whatever tags we use) to an indented approach, to hightlight it better.

This also gives us better flexibility when we start adding some CSS styling to the HTML rendering, where the indented variant rendering gives much better abilities the formatting and styling.  Essentially it changes from a more plain `<p>...</p>`  for "everything" to a `<dl><dt>$HEADER</dt><dd>$CONTENT</dd></dl>` style.

ordex: should we keep this patch as is and then reformat all the NOTEs with a follow up?

ordex: @mattock would you mind rebasing this patch on master and sending it to the mailing list? Once that's done, we can close this PR. Thanks!

PR #176, opened 2022-07-02
To be able to run OpenVPN client in Docker container

mtilson:
The file `/etc/resolv.conf` is not a regular one in Docker container, it is a bind-mount within container.
It cause `cp: can't create '/etc/resolv.conf': File exists` error and as a result `resolv.conf.ovpnsave` is not copied back as it is intended.
But rewriting the file instead of moving or renaming is possible.

### DIFF
--- a/contrib/pull-resolv-conf/client.down
+++ b/contrib/pull-resolv-conf/client.down
@@ -44,7 +44,7 @@ if type resolvconf >/dev/null 2>&1; then
   resolvconf -d "${dev}" -f
 elif [ -e /etc/resolv.conf.ovpnsave ] ; then
   # cp + rm rather than mv in case it's a symlink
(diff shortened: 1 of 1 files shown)


ordex:
I think this makes sense. @cron2 @dsommers do you have any opinion?

In any case, a proper patch should be eventually sent to the mailing list.

dsommers:
this is a contrib script; normally not installed in any default $PATH.  I consider these contrib scripts more like templates.  Change itself is fine for me.

ordex:
@mtilson can we have your SoB line on the patch (preferably sent to the mailing list)?
Otherwise it cannot be processed.

ordex:
@mtilson any chance you could address the comments above?

Gelassen:
@mtilson , @ordex , any news on this?  I am looking for a reliable docker image to deploy on my VPS and haven't so far the right one.

PR #178, opened 2022-09-10
Fix typo in "Update android.txt"

mtwalli:
Fix


cron2:
Thanks for your contribution.

I have used that opportunity to ask Arne to review the whole document, and he sent a patch with many more changes (rewriting the paragraph you found in the process).  So this PR won't apply anymore, but served a useful purpose.

PR #278, opened 2023-03-15
update to OpenSSL3 semantics of EVP_PKEY_id

baentsch:
Before submitting the patch to the mailing list this is to document the problem this fixes:

Beginning with OpenSSL3.0.0, the function `EVP_PKEY_id` added the possible return value -1 / [EVP_PKEY_KEYMGMT](https://github.com/openssl/openssl/blob/67bfdfa17bc4ca8e2b819316299bb82748394c45/include/openssl/evp.h#L82-L83) indicating the PKEY as being implemented in a [provider](https://www.openssl.org/docs/manmaster/man7/provider.html). This value is currently not properly tested for [in OpenVPN here](https://github.com/OpenVPN/openvpn/blob/021952705ba4b77a1d94bec85f6f02e6c612a37e/src/openvpn/ssl_openssl.c#L2083-L2086), leading to the case where OpenVPN stops working properly when triggering this code, e.g., when activating PQ crypto (via [oqs-provider](https://github.com/open-quantum-safe/oqs-provider)) together with "verbosity" set to 5: `OBJ_nid2sn(-1)` is an error. For further background, see https://github.com/open-quantum-safe/oqs-demos/issues/183 and https://github.com/openssl/openssl/pull/20501.

### DIFF
--- a/src/openvpn/ssl_openssl.c
+++ b/src/openvpn/ssl_openssl.c
@@ -1501,7 +1501,11 @@ tls_ctx_use_management_external_key(struct tls_root_ctx *ctx)
     }
     EVP_PKEY_free(privkey);
 #else  /* ifdef HAVE_XKEY_PROVIDER */
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
     if (EVP_PKEY_id(pkey) == EVP_PKEY_RSA)
+#else /* OPENSSL_VERSION_NUMBER < 0x30000000L */
+    if (EVP_PKEY_is_a(pkey, "RSA"))
+#endif /* OPENSSL_VERSION_NUMBER < 0x30000000L */
     {
         if (!tls_ctx_use_external_rsa_key(ctx, pkey))
         {
@@ -1509,7 +1513,11 @@ tls_ctx_use_management_external_key(struct tls_root_ctx *ctx)
         }
     }
 #if (OPENSSL_VERSION_NUMBER > 0x10100000L) && !defined(OPENSSL_NO_EC)
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
     else if (EVP_PKEY_id(pkey) == EVP_PKEY_EC)
+#else /* OPENSSL_VERSION_NUMBER < 0x30000000L */
+    else if (EVP_PKEY_is_a(pkey, "EC"))
+#endif /* OPENSSL_VERSION_NUMBER < 0x30000000L */
     {
         if (!tls_ctx_use_external_ec_key(ctx, pkey))
         {
@@ -2064,10 +2072,15 @@ print_cert_details(X509 *cert, char *buf, size_t buflen)
     }
 
     int typeid = EVP_PKEY_id(pkey);
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
+    bool is_ec = typeid == EVP_PKEY_EC;
+#else
+    bool is_ec = EVP_PKEY_is_a(pkey, "EC");
+#endif
 
 #ifndef OPENSSL_NO_EC
     char groupname[256];
-    if (typeid == EVP_PKEY_EC)
+    if (is_ec)
     {
         size_t len;
         if (EVP_PKEY_get_group_name(pkey, groupname, sizeof(groupname), &len))
@@ -2080,9 +2093,9 @@ print_cert_details(X509 *cert, char *buf, size_t buflen)
         }
     }
 #endif
-    if (EVP_PKEY_id(pkey) != 0)
+    if (typeid != 0)
     {
-        int typeid = EVP_PKEY_id(pkey);
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
         type = OBJ_nid2sn(typeid);
 
         /* OpenSSL reports rsaEncryption, dsaEncryption and
@@ -2104,6 +2117,13 @@ print_cert_details(X509 *cert, char *buf, size_t buflen)
         {
             type = "unknown type";
         }
+#else /* OpenSSL >= 3 */
+        type = EVP_PKEY_get0_type_name(pkey);
+        if (type == NULL)
+        {
+            type = "(error getting public key type)";
+        }
+#endif /* if OPENSSL_VERSION_NUMBER < 0x30000000L */
     }
 
     char sig[128] = { 0 };


schwabe:
Do you know what EVP_PKEY_get0_type_name outputs for common key keytypes that we currenlty have our custom code? Since this is only informal message if the output is reasonable enough, we might be able to always use EVP_PKEY_get0_type_name instead of our own custom key -> name function.

There seem to be other usages of EVP_PKEY_id.

I am a bit surprised about OpenVPN stopping to work as this should only trigger `type = "unknown type";` or does `OBJ_nid2sn(-1)` mess things up?

baentsch:
> Do you know what EVP_PKEY_get0_type_name outputs for common key keytypes that we currenlty have our custom code?

No -- to limit the amount of code change(d) I only tested `EVP_PKEY_get0_type_name` with provider-based algorithms (triggering the -1 response). But I'd be glad to change the patch and let you know. The old code is still needed for OpenSSL<3, right? In any case using `EVP_PKEY_get0_type_name` with an OpenSSL3 integration however is the recommendation of the OpenSSL core team (https://github.com/openssl/openssl/pull/20501#issuecomment-1467573388)

> does OBJ_nid2sn(-1) mess things up?

Precisely that: It triggers an error condition in OpenSSL that leads to problems later on.

schwabe:
Yes the old cod will be still needed for < OpenSSL 3.0 but having it all in a < 3.0 ifdef will allow us to eventually get rid of entirely and only the new recommended API.

baentsch:
ACK. Will amend the PR (with a "reverse" version check (< 3.0.0)).

Regarding the CI failure, is there anything where I can easily check and see how to run `uncrustify` (personally unfamiliar with the tool) to fix the problem reported?

cron2:
Hi,

On Wed, Mar 15, 2023 at 07:58:37AM -0700, Michael Baentsch wrote:
> Regarding the CI failure, is there anything where I can easily check and see how to run `uncrustify` (personally unfamiliar with the tool) to fix the problem reported?

The easiest way might be to use the uncrustify pre-commit-hook (provided
you have uncrustify installed), by running

  dev-tools/git-pre-commit-uncrustify.sh install

from within your openvpn git clone - this will run uncrustify on every
commit, with the rules in dev-tools/uncrustify.conf, and complain (with
a diff) if it's seen as "non compliant".

Unfortunately there's a catch - uncrustify code base is not really stable,
so "most recent" versions produce erroneous output.  We use 0.71 or 0.72
and these seem to work reliably.

gert

-- 

baentsch:
@schwabe : 

> Do you know what EVP_PKEY_get0_type_name outputs for common key keytypes that we currenlty have our custom code?

After updating the PR to generally use `EVP_PKEY_get0_type_name` I now ran my harness with "rsa" and "ed448" as classic sig algs and the generated log output at verbosity 5 now reads "Control Channel: TLSv1.3, cipher TLSv1.3 TLS_AES_256_GCM_SHA384, peer certificate: 2048 bit RSA, signature: RSA-SHA256" and "Control Channel: TLSv1.3, cipher TLSv1.3 TLS_AES_256_GCM_SHA384, peer certificate: 456 bit ED448, signature: ED448", respectively. When running with one sample PQ/provider algorithm, it reads "Control Channel: TLSv1.3, cipher TLSv1.3 TLS_AES_256_GCM_SHA384, peer certificate: 256 bit falcon1024, signature: falcon1024"

@selvanair : Thanks very much for the `uncrustify` HowTo. Worked like a charm. Let' see how the updated PR does in CI...

baentsch:
Oops -- I did the PR update using my normal "anonymous" github email-address -- did this stop CI from running? Would someone else want to trigger CI or should I re-do the latest commit using my "real" email address? Or squash both commits into one using that address? Or just submit an email with the patch to the mailing list?

cron2:
Hi,

On Wed, Mar 15, 2023 at 10:08:22AM -0700, Michael Baentsch wrote:
> Oops -- I did the PR update using my normal "anonymous" github email-address -- did this stop CI from running? Would someone else want to trigger CI or should I re-do the latest commit using my "real" email address? Or squash both commits into one using that address? Or just submit an email with the patch to the mailing list?

The final step ("submit an e-mail with the squashed patch to the list")
is required anyway.  Since the uncrustify complaints have beem minor,
I'll just take care at merge time if there is anything left.

(Core developers are expected to get it right by running the pre-commit
hook locally, other contributors get more liberal treatment ;-) )

gert

-- 

baentsch:
@cron2 Thanks for the guidance. Squash done. Please let me know if things look OK here to run `git send-email --to=openvpn-devel@lists.sourceforge.net HEAD~1`.

src/openvpn/ssl_openssl.c
```diff
@@ -2104,6 +2104,13 @@ print_cert_details(X509 *cert, char *buf, size_t buflen)
         {
             type = "unknown type";
         }
+#else /* OpenSSL >= 3 */
+        type = EVP_PKEY_get0_type_name(pkey);
+        if (type == NULL)
+        {
+            type = "(error getting OpenSSL3 type)";
```
levitte, line 2111: If I saw that as a user, I would not understand it at all.  May I suggest something like `"(error getting public key type)"`?

levitte:
Just above the section you change, I detect an old way of detecting an EC key to get the group.  I would apply this diff there:

``` diff
 #ifndef OPENSSL_NO_EC
     char groupname[256];
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
+    int is_ec = typeid == EVP_PKEY_EC;
+#else
+    int is_ec = EVP_PKEY_is_a(pkey, "EC");
+#endif
-    if (typeid == EVP_PKEY_EC)
+    if (is_ec)
     {
         size_t len;
         if (EVP_PKEY_get_group_name(pkey, groupname, sizeof(groupname), &len))
         {
             curve = groupname;
         }
         else
         {
             curve = "(error getting curve name)";
         }
     }
 #endif
```

schwabe:
we use C99 and bool in OpenVPN but otherwise I agree with using EVP_PKEY_is_a

baentsch:
> Just above the section you change, I detect an old way of detecting an EC key to get the group. I would apply this diff there 

Will do. And there's more: https://github.com/OpenVPN/openvpn/blob/e8ecaadd2ac38f2c2d4bcd40eeaea7401aa737a1/src/openvpn/ssl_openssl.c#L1504 and https://github.com/OpenVPN/openvpn/blob/e8ecaadd2ac38f2c2d4bcd40eeaea7401aa737a1/src/openvpn/ssl_openssl.c#L1512

src/openvpn/ssl_openssl.c
```diff
@@ -1501,15 +1501,23 @@ tls_ctx_use_management_external_key(struct tls_root_ctx *ctx)
...
         {
             goto cleanup;
         }
     }
 #if (OPENSSL_VERSION_NUMBER > 0x10100000L) && !defined(OPENSSL_NO_EC)
+#if OPENSSL_VERSION_NUMBER < 0x30000000L
     else if (EVP_PKEY_id(pkey) == EVP_PKEY_EC)
+#else /* OPENSSL_VERSION_NUMBER < 0x30000000L */
+    else if (EVP_PKEY_is_a(pkey, "EC"))
+#endif /* OPENSSL_VERSION_NUMBER < 0x30000000L */
```
schwabe, line 1520: These two chunks might not be really that important. They are inside the block if we do not have `HAVE_XKEY_PROVIDER`. Then instead of implementing external key usage by using a provider (xkey) we use EC_KEY_METHOD and RSA_KEY_METHOD method and a ton of other deprecated APIs.

schwabe: I don't have a strong preference. You could add a note that this code path is only used when we don't use xkey provider, ie only when the depracted *_KEY_METHOD are used. But I will accept the patch either way.

baentsch:
Closing here as PR submitted to mailing list

PR #351, opened 2023-06-11
optimize file info when start openvpn server

ropodaing:

### DIFF
--- a/src/openvpn/options.c
+++ b/src/openvpn/options.c
@@ -3888,7 +3888,7 @@ check_file_access(const int type, const char *file, const int mode, const char *
     /* Scream if an error is found */
     if (errcode > 0)
     {
-        msg(M_NOPREFIX | M_OPTERR | M_ERRNO, "%s fails with '%s'", opt, file);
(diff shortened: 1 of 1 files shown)


src/openvpn/options.c
```diff
@@ -3888,7 +3888,7 @@ check_file_access(const int type, const char *file, const int mode, const char *
     /* Scream if an error is found */
     if (errcode > 0)
     {
-        msg(M_NOPREFIX | M_OPTERR | M_ERRNO, "%s fails with '%s'", opt, file);
+        msg(M_NOPREFIX | M_OPTERR | M_ERRNO, "%s fails with '%s', you can check file Permissions or selinux security context", opt, file);
```
ordex, line 3891: How can you be sure that the error is about permissions? It may also be something totally unrelated.

schwabe: a read error or broken device or broken symlink will also trigger this. Also it is missing `"`.

cron2:
I do not think we should merge this - the error message already prints what it doesn't like ("permission denied", "no such file or directory", etc.), so the longer message is not adding much useful information for the "permission denied" case, and adds confusion for all other cases.

Issue #328, opened 2023-05-05
Openvpn 2.6.3 cras "nregister_netdevice: waiting for tun_0 to become free. Usage count = 1"


kolya25:
**Describe the bug**
Аfter restarting the openvpn service, it gives an error, the server does not restart, only hard reset works.
```
Message from syslogd@vpn at May  5 14:20:48 ...
 kernel:[83842.337867] unregister_netdevice: waiting for tun_0 to become free. Usage count = 1
```
**To Reproduce**
service openvpn restart or reboot system

**Version information :**
 - OS: Debian 10, kernel 5.10.0-21-amd64 
 - OpenVPN version: 2.6.3 x86_64-pc-linux-gnu
 - DCO version: 0.2.20230323-5-g961c60d

ordex:
Hi, this is a duplicate of https://github.com/OpenVPN/ovpn-dco/issues/18
Do you have any reliable way to reproduce this issue?

Issue #267, opened 2023-03-02
SOCKS proxy 5 second handshake/auth/connect hard-coded timeout


ValdikSS:
**Describe the bug**
OpenVPN 2.5, 2.6 and master have a hard-coded timeout of 5 seconds for SOCKS communications, which may not be enough for slow proxies, such as when the connection is performed over Tor.

**To Reproduce**
1. Setup OpenVPN server using TCP
2. Configure Tor Hidden Service to that server
3. On another machine, configure client to Tor hidden service VPN using Tor's SOCKS port `socks-proxy 127.0.0.1 9050`

**Expected behavior**
Proxy timeout follow `--server-poll-timeout` value
>--server-poll-timeout n
>    When  connecting  to  a remote server do not wait for more than n seconds for a response before trying the next server. The de‐fault value is 120s. **This timeout includes proxy and TCP connect timeouts**.

**Actual behavior**
Socks code has a 5 second hard-coded timeout in numerous places
https://github.com/OpenVPN/openvpn/blob/9c52e0c610ef1229561c2d038ca41fe2cbefe8da/src/openvpn/socks.c#L95
https://github.com/OpenVPN/openvpn/blob/9c52e0c610ef1229561c2d038ca41fe2cbefe8da/src/openvpn/socks.c#L194
https://github.com/OpenVPN/openvpn/blob/9c52e0c610ef1229561c2d038ca41fe2cbefe8da/src/openvpn/socks.c#L312

ValdikSS:
I can confirm that the patch works great.

PR #411, opened 2023-09-15
Implement server_poll_timeout for socks instead of hard-coded 5 sec

5andr0:
People have been struggling to run openvpn over slow proxies. 
There was a bug [#328](https://community.openvpn.net/openvpn/ticket/328) submitted 10 years ago. 
To fix this, the server-poll-timeout config value was introduced, but has only been implemented for http proxies. 
For socks the timeouts are still hard-coded to 5 sec.
That's why people are still having problems, see [issue #267](https://github.com/OpenVPN/openvpn/issues/267) 

I submitted a [patch](https://www.mail-archive.com/openvpn-devel@lists.sourceforge.net/msg26962.html) to the mailing list, but since this was my first time using a mailing list over git, I failed to provide details in the mail.

With this patch, establishing a socks connection uses the timeout set by server-poll-timeout (default: 120s).
I used the same logic like it was implemented for http proxies, which includes the socks handshake for the timeout.

So here's a PR with more details, in hope that it gets reviewed and implemented soon.

### DIFF
--- a/src/openvpn/socket.c
+++ b/src/openvpn/socket.c
@@ -2075,6 +2075,7 @@ phase2_tcp_client(struct link_socket *sock, struct signal_info *sig_info)
                                            sock->sd,
                                            sock->proxy_dest_host,
                                            sock->proxy_dest_port,
+                                           sock->server_poll_timeout,
                                            sig_info);
         }
         if (proxy_retry)
@@ -2104,6 +2105,7 @@ phase2_socks_client(struct link_socket *sock, struct signal_info *sig_info)
                                    sock->ctrl_sd,
                                    sock->sd,
                                    &sock->socks_relay.dest,
+                                   sock->server_poll_timeout,
                                    sig_info);
 
     if (sig_info->signal_received)
--- a/src/openvpn/socks.c
+++ b/src/openvpn/socks.c
@@ -42,6 +42,7 @@
 #include "fdmisc.h"
 #include "misc.h"
 #include "proxy.h"
+#include "forward.h"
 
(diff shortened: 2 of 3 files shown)


5andr0:
> A commit needs to have a good description on what is changed and why - there is no difference between "PR" and "on the list" in this. Your commit has a summary, but the description that is here in the PR ("the same logic that was implemented for http proxies") needs to go into the actual commit message - when people look at code change some years from now, github might long be gone (so whatever is written in the PR comments is long history), but the actual git commit message will be still there.

I highly appreciate you taking the time to explain how it's done properly. From now on I'll always write multi-line commit messages. 

> Also, this PR has a merge commit, which it shouldn't have (so always rebase on master before creating a patch).

Rebased and amended a detailed commit message 👍🏻

> ... but do not despair, we do have the patch on the list, and with the description from the PR, I can construct a more verbose commit message. I just want a review, preferrably from @schwabe who understands all this timeout stuff (having rewritten much of it a few years ago)...

Thanks for taking over!

ValdikSS:
@cron2
LGTM, tested on 2.5.8

cron2:
Thanks for testing.  By convention, we never merge GH PRs, but merge patches from the mailing list.

This has now been done

commit b3a68b85a729628ca8b97f9f0c2813f795289cfc (master)
commit 94bfb712366ece1ca3605d18e99580f482f0232b (release/2.6)
Author: 5andr0
Date:   Fri Mar 15 17:20:11 2024 +0100

     Implement server_poll_timeout for socks

- so, thanks to all involved.

PR #441, opened 2023-10-26
protocol_dump: tls-crypt support

### LINKED ISSUE #440: Protocol_dump incorrect for tls-crypt packets
It seems the flags set for `tls_auth_hmac_size` are not used for tls-crypt first of all (if they were it would not be correct as the hmac and replay packet id are swapped between tls-auth and tls-crypt packets). Then the ack list is read incorrectly: the most significant bit of the replay packet id is read as the ack list length. If the client doesn't support hmac cookies this value is likely 0. In that case bytes 1..5 of the replay packet id is read as the packet id (message id). Then the remainder of the header and the data is printed as data.
https://github.com/OpenVPN/openvpn/blob/c827f9d83a7246971f435d0053b0252e49770f11/src/openvpn/ssl.c#L4180-L4227

It's not clear to me how to fix this, but likely a flag needs to be added and passed for tls-crypt.

reynir:
Fixes #440

There are perhaps some silly choices to show the ACK list as `[ENCRYPTED]` etc.

### DIFF
--- a/src/openvpn/openvpn.h
+++ b/src/openvpn/openvpn.h
@@ -544,7 +544,8 @@ struct context
 #define PROTO_DUMP(buf, gc) protocol_dump((buf), \
                                           PROTO_DUMP_FLAGS   \
                                           |(c->c2.tls_multi ? PD_TLS : 0)   \
-                                          |(c->options.tls_auth_file ? md_kt_size(c->c1.ks.key_type.digest) : 0), \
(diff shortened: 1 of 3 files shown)


cron2:
commit 227799b8345128dd3adf2029323457804209fe93
Author: Reynir Björnsson <reynir@reynir.dk>
Date:   Thu Oct 26 16:55:32 2023 +0200

    protocol_dump: tls-crypt support

(via the openvpn-devel@ mailing list)

PR #463, opened 2023-11-27
Fix send_reliable array index in check_session_buf_not_used()

braindead-bf:
A for loop within `check_session_buf_not_used()` uses `j` as an iterator, then checks the `send_reliable` array at index `i`. It should check at index `j` instead.


schwabe:
There is already a patch pending for the thing you found: https://gerrit.openvpn.net/c/openvpn/+/459

braindead-bf:
Great, thanks!

PR #466, opened 2023-12-04
bump pkcs11-helper to version 1.30

astos-marcb:
Some local fixes were incorporated in upstream and can be removed.

As anticipated, the DLL path feature landed during OpenVPN 2.6 lifetime.
The build fix by @flichtenheld is likely not end-user-visible.


cron2:
Patch is on the list, and is being merged right now...

commit a78b0e45dff3a0f0332de47c55aadd76c5919370 (master)
commit 77b2e940fd3980f9dbfa62f84fb135ea2f1effaf (release/2.6)

PR #484, opened 2024-01-07
Update README.dco.md

codesmith-emmy:

### DIFF
--- a/README.dco.md
+++ b/README.dco.md
@@ -93,15 +93,15 @@ configuring routes to clients (aka "iroutes") and consults the main kernel
 routing tables for forwarding decisions.
 
 - Each client has a VPN IPv4 and/or a VPN IPv6 assigned to it;
-- additional IP ranges can be routed to a client by adding a route with
+- Additional IP ranges can be routed to a client by adding a route with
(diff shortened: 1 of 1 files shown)


README.dco.md
```diff
@@ -93,33 +93,33 @@ configuring routes to clients (aka "iroutes") and consults the main kernel
 routing tables for forwarding decisions.
 
 - Each client has a VPN IPv4 and/or a VPN IPv6 assigned to it;
-- additional IP ranges can be routed to a client by adding a route with
+- Additional IP ranges can be routed to a client by adding a route with
```
ordex, line 96: To the best of my knowledge semi-color does not require the following sentence to start with capital letter.
Do you have any reference claiming the opposite?

codesmith-emmy: You are very correct. However, I didn’t suggest the changes because the sentence is wrong. My goal was to prevent anyone who may be skimming through the README from mixing things up.
I’m okay with any decision regarding my suggested changes as this is not a big issue.

ordex: I can't see what someone may mix up. And if you agree that the change is not required..why proposing it in the first place?

PR #550, opened 2024-05-07
support freebsd's fib feature

cnbatch:
This patch is to support FreeBSD's FIB feature.

When using FreeBSD's `setfib` command to launch OpenVPN in other FIB (routing table), the interface's IP address was not added to correct routing table, and then the routing entries couldn't be added successfully:
https://forums.freebsd.org/threads/freebsd-14-and-route-in-non-zero-fib.91099/
https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=278295

This is duo to `ifconfig` command does not retrieve FIB number of current envirnment which is set by `setfib`. It only accepts `fib N` parameter.
`route` command can retrieve FIB number of current envirnment, the following routing entries will be added to correct routing table. But if the interface's IP address was not added correct routing table, `route` command will fail.

### DIFF
--- a/src/openvpn/tun.c
+++ b/src/openvpn/tun.c
@@ -55,6 +55,11 @@
 
 #include <string.h>
 
+#ifdef TARGET_FREEBSD
+#include <sys/types.h>
+#include <sys/sysctl.h>
+#endif
+
 #ifdef _WIN32
 
 const static GUID GUID_DEVCLASS_NET = { 0x4d36e972L, 0xe325, 0x11ce, { 0xbf, 0xc1, 0x08, 0x00, 0x2b, 0xe1, 0x03, 0x18 } };
@@ -1169,14 +1174,43 @@ do_ifconfig_ipv6(struct tuntap *tt, const char *ifname, int tun_mtu,
         openvpn_execve_check(&argv, es, 0, "Solaris ifconfig IPv6 mtu failed");
     }
 #elif defined(TARGET_OPENBSD) || defined(TARGET_NETBSD) \
-    || defined(TARGET_DARWIN) || defined(TARGET_FREEBSD) \
-    || defined(TARGET_DRAGONFLY)
+    || defined(TARGET_DARWIN) || defined(TARGET_DRAGONFLY)
     argv_printf(&argv, "%s %s inet6 %s/%d mtu %d up", IFCONFIG_PATH, ifname,
                 ifconfig_ipv6_local, tt->netbits_ipv6, tun_mtu);
     argv_msg(M_INFO, &argv);
 
     openvpn_execve_check(&argv, es, S_FATAL,
                          "generic BSD ifconfig inet6 failed");
+
+#elif defined(TARGET_FREEBSD)
+    /* read current fib number, codes are from: */
+    /* https://github.com/freebsd/freebsd-src/blob/f9716eee8ab45ad906d9b5c5233ca20c10226ca7/sbin/route/route.c#L269 */
+    int numfibs = 0, defaultfib = 0;
+	size_t len = sizeof(numfibs);
+	if (sysctlbyname("net.fibs", (void *)&numfibs, &len, NULL, 0) == -1)
+		numfibs = -1;
+
+	len = sizeof(defaultfib);
+	if (numfibs != -1 &&
+	    sysctlbyname("net.my_fibnum", (void *)&defaultfib, &len, NULL,
+		0) == -1)
+		defaultfib = -1;
+
+    if (defaultfib <= 0)
+    {
+        argv_printf(&argv, "%s %s inet6 %s/%d mtu %d up", IFCONFIG_PATH, ifname,
+                    ifconfig_ipv6_local, tt->netbits_ipv6, tun_mtu);
+    }
+    else
+    {
+        argv_printf(&argv, "%s %s inet6 %s/%d mtu %d up fib %d", IFCONFIG_PATH, ifname,
+                    ifconfig_ipv6_local, tt->netbits_ipv6, tun_mtu, defaultfib);
+    }
+
+    argv_msg(M_INFO, &argv);
+
+    openvpn_execve_check(&argv, es, S_FATAL,
+                         "FreeBSD ifconfig inet6 failed");
 
 #if defined(TARGET_FREEBSD) && __FreeBSD_version >= 1200000 \
     && __FreeBSD_version < 1300000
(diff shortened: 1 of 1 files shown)


src/openvpn/tun.c
```diff
@@ -1169,15 +1174,44 @@ do_ifconfig_ipv6(struct tuntap *tt, const char *ifname, int tun_mtu,
...
-    || defined(TARGET_DRAGONFLY)
+    || defined(TARGET_DARWIN) || defined(TARGET_DRAGONFLY)
     argv_printf(&argv, "%s %s inet6 %s/%d mtu %d up", IFCONFIG_PATH, ifname,
                 ifconfig_ipv6_local, tt->netbits_ipv6, tun_mtu);
     argv_msg(M_INFO, &argv);
 
     openvpn_execve_check(&argv, es, S_FATAL,
                          "generic BSD ifconfig inet6 failed");
 
+#elif defined(TARGET_FREEBSD)
```
flichtenheld, line 1185: There is code further below (line 1239) that you would need to move up.
```suggestion
#if defined(TARGET_OPENBSD) || defined(TARGET_NETBSD) \
    || defined(TARGET_DARWIN)
    /* and, hooray, we explicitly need to add a route... */
    add_route_connected_v6_net(tt, es);
#endif

#elif defined(TARGET_FREEBSD)
```

src/openvpn/tun.c
```diff
@@ -1573,6 +1607,55 @@ do_ifconfig_ipv4(struct tuntap *tt, const char *ifname, int tun_mtu,
...
 
+    argv_msg(M_INFO, &argv);
+    openvpn_execve_check(&argv, es, S_FATAL, "DragonflyBSD ifconfig failed");
+
+#elif defined(TARGET_FREEBSD)
+
+    /* read current fib number, codes are from: */
+    /* https://github.com/freebsd/freebsd-src/blob/f9716eee8ab45ad906d9b5c5233ca20c10226ca7/sbin/route/route.c#L269 */
+    int numfibs = 0, defaultfib = 0;
+	size_t len = sizeof(numfibs);
```
flichtenheld, line 1618: Wrong indentation

src/openvpn/tun.c
```diff
@@ -1169,15 +1174,44 @@ do_ifconfig_ipv6(struct tuntap *tt, const char *ifname, int tun_mtu,
...
     argv_msg(M_INFO, &argv);
 
     openvpn_execve_check(&argv, es, S_FATAL,
                          "generic BSD ifconfig inet6 failed");
 
+#elif defined(TARGET_FREEBSD)
+    /* read current fib number, codes are from: */
+    /* https://github.com/freebsd/freebsd-src/blob/f9716eee8ab45ad906d9b5c5233ca20c10226ca7/sbin/route/route.c#L269 */
+    int numfibs = 0, defaultfib = 0;
+	size_t len = sizeof(numfibs);
```
flichtenheld, line 1189: Wrong indentation

flichtenheld (changes requested):

schwabe:
I am wondewring if the ifconfig behaviour is intended behaviour or if we working around a bug in ifconfig here in OpenVPN.

schwabe:
There are also build failures related to macOS.

flichtenheld:
> There are also build failures related to macOS.

Yeah, these are related to the code that was not moved correctly.

src/openvpn/tun.c
```diff
@@ -1169,15 +1174,50 @@ do_ifconfig_ipv6(struct tuntap *tt, const char *ifname, int tun_mtu,
...
-    || defined(TARGET_DRAGONFLY)
+    || defined(TARGET_DARWIN) || defined(TARGET_DRAGONFLY)
     argv_printf(&argv, "%s %s inet6 %s/%d mtu %d up", IFCONFIG_PATH, ifname,
                 ifconfig_ipv6_local, tt->netbits_ipv6, tun_mtu);
     argv_msg(M_INFO, &argv);
 
     openvpn_execve_check(&argv, es, S_FATAL,
                          "generic BSD ifconfig inet6 failed");
 
+#if defined(TARGET_OPENBSD) || defined(TARGET_NETBSD) \
```
flichtenheld: You need to move the code, not copy it. But I can't suggest that part via the GitHub UI since that is too far outside the code changes ;)

src/openvpn/tun.c
cnbatch: truly sorry for my typos ::>_<::

cnbatch (commented):
all changes have made

cnbatch:
Sorry, it seems I misunderstood earlier.
This doesn't appear to be a FreeBSD bug but rather a requirement of `ifconfig`. According to FreeBSD's documentation, if a FIB number needs to be specified, it must be explicitly stated it in `ifconfig`; otherwise, the default FIB (main route table) will be used:
https://man.freebsd.org/cgi/man.cgi?ifconfig

```
fib	fib_number
	       Specify interface FIB.  A FIB fib_number	 is  assigned  to  all
	       frames  or  packets received on that interface.	The FIB	is not
	       inherited, e.g.,	vlans or other sub-interfaces will use the de-
	       fault FIB (0) irrespective of the parent	interface's FIB.   The
	       kernel  needs  to be tuned to support more than the default FIB
	       using the  ROUTETABLES  kernel  configuration  option,  or  the
	       net.fibs	tunable.
```

janondrusek:
Is there a way this could be rebased and merged? FreeBSD still cannot effectively use a separate routing table with OpenVPN.

My own attempts end with errors too:

```
2025-10-13 20:35:50 /sbin/ifconfig tun0 10.x.x.x/16 mtu 1500 up
add net y.y.y.y: gateway 127.0.0.1 fib 1
add net 0.0.0.0: gateway 10.x.x.1 fib 1: Invalid argument
2025-10-13 20:35:50 ERROR: FreeBSD route add command failed: external program exited with error status: 1
add net 128.0.0.0: gateway 10.x.x.1 fib 1: Invalid argument
2025-10-13 20:35:50 ERROR: FreeBSD route add command failed: external program exited with error status: 1
2025-10-13 20:35:50 Initialization Sequence Completed
```

cron2:
This is a relevant feature, but needs good design and careful thinking (OpenBSD has FIB support as well, Linux has vrf/namespace support, and we want to be able to ifconfig in whatever a platform has).

For 2.7, the ship has sailed - lack of time on my side to followup on this, and do more thinking on the design.  2.7.0 release is "around the corner", and we're in the "bugfixes, integers, and non-intrusive last-minute changes" territory.

We need to come back to this after 2.7.0 release, so roughly "end of November".

cron2:
Side note: `networking_freebsd.c` already has fib support, we just don't use it for normal routes yet.  This is one of the places where a larger cleanup is needed anyway, and we get route-fib-support for free.  Won't help with ifconfig-fib-support.

PR #554, opened 2024-05-16
fix error "UDPv6: Can't assign requested address (code=49)"

Tontonjo:
This bug appears when you let the openvpn server in "multihome" configuration, preventing the connexion to be established.

Related to and fixed by user hashiz in openvpn community but not yet implemented.

https://community.openvpn.net/openvpn/attachment/ticket/874/patch-src-openvpn-socket.c

Tested this fix on my setup with success

### DIFF
--- a/src/openvpn/socket.c
+++ b/src/openvpn/socket.c
@@ -3481,7 +3481,7 @@ link_socket_write_udp_posix_sendmsg(struct link_socket *sock,
             cmsg->cmsg_type = IPV6_PKTINFO;
 
             pkti6 = (struct in6_pktinfo *) CMSG_DATA(cmsg);
-            pkti6->ipi6_ifindex = to->pi.in6.ipi6_ifindex;
+            pkti6->ipi6_ifindex = 0;
             pkti6->ipi6_addr = to->pi.in6.ipi6_addr;
             break;
         }


schwabe:
We assume that you can reply on the same interface (ifindex) that you get message in. You are removing that assumption. I think your change will break talking to clients that connect via link local addresses as for those you need to specify the ifindex to be able to reply. 

So it would be good to have a better understanding of your environment and why this is happening. Do you have some assymetric routing setup?

Tontonjo:
> We assume that you can reply on the same interface (ifindex) that you get message in. You are removing that assumption. I think your change will break talking to clients that connect via link local addresses as for those you need to specify the ifindex to be able to reply.
> 
> So it would be good to have a better understanding of your environment and why this is happening. Do you have some assymetric routing setup?

As far as i know i dont have an asymetric setup.

Let me clarify a bit:
This fixed my problem where i was able to correctly establish connexion over ipv6 in "multihome" mode.

My VPN would correctly be established if i'd set the "local" address to my ipv6 interface address, but this would imply to lose ipv4 connectivity, which i dont want to.

This fix is not mine but as far as the bug seems to have been stalled and this indeed fixed it in my case, i though it would be cool to contribute.

Regards

cron2:
Mmmh, this seems to be a duplicate of https://github.com/OpenVPN/openvpn/pull/856 which has a much more helpful subject line...

And no, this hasn't been fixed, because the "simple fix" breaks other scenarios.

Tontonjo:
> Mmmh, this seems to be a duplicate of #856 which has a much more helpful subject line...
> 
> And no, this hasn't been fixed, because the "simple fix" breaks other scenarios.

What brought you there, my dear, after all this time?

cron2:
> > Mmmh, this seems to be a duplicate of #856 which has a much more helpful subject line...
> > And no, this hasn't been fixed, because the "simple fix" breaks other scenarios.
> 
> What brought you there, my dear, after all this time?

Cleaning up GH PRs.  Since we never merge PRs, and discourage them for everything except initial review, these are by definition all stale - but there are over 40 open PRs, which is not how things should be.

Tontonjo:
> > > Mmmh, this seems to be a duplicate of #856 which has a much more helpful subject line...
> > > And no, this hasn't been fixed, because the "simple fix" breaks other scenarios.
> > 
> > 
> > What brought you there, my dear, after all this time?
> 
> Cleaning up GH PRs. Since we never merge PRs, and discourage them for everything except initial review, these are by definition all stale - but there are over 40 open PRs, which is not how things should be.

For sure it should have been closed by the time.

cron2:
So, the behaviour will change in 2.7_rc3 and subsequently in 2.7.0 - see the linked commit, which has been merged just now.

PR #591, opened 2024-08-14
RFC: Interpret --key and --cert option argument as URI

selvanair:
OpenSSL 3 has providers which can load keys and certificates from various key stores and HSMs using a provider-specific URI. While certificates are generally exportable, and some providers support a PEM file that acts as a proxy for non-exportable private keys, not all providers are expected to do so. A generic capability to read keys and certificates from URIs appears useful.

This patch does this by extending the scope of the argument for "--key" and "--cert" options to include URIs. Many of OpenSSL 3 utilities also work the same way: e.g., the "-in" option for "openssl pkey" or "openssl x509" could be a filename or URI. Other applications have started emulating this behaviour: e.g., pkcs11: URI works as an alternative to a file name for certificates and keys in apache. Even for files, this has a nice side effect that non-PEM files get transparently parsed. E.g., a pkcs12 file could be used in place of a PEM file without needing any extra options.

This is backward compatible as OpenSSL falls back to treating URIs with no scheme or unrecognized scheme as file names.

Parsing of inlined keys and certificates is unchanged (those should be in PEM format).

Specification of URIs that OpenSSL accepts depends on the providers that support them. Some are standard URIs such as "file:/path", but providers may support non-standard URIs with arbitrary scheme names. OpenSSL by itself recognizes only file URI.  However, the implementation is agnostic to the URI specification as parsing is done by the provider that supports the URI. A new URI gets automatically recognized when the provider that supports it is loaded.

Below are some usage examples:

Relative or absolute path to a file or as a URI "file:/absolute/path":

   --key mykey.pem      (same as what is currently supported)
   --key file:/path/to/mykey.pem
   --cert file:/path/to/mycert.pem

Other file types supported by OpenSSL would also work:

   --key client.p12
   --cert client.p12

pkcs11-provider supports "pkcs11:" URI (RFC 7512):

   --key pkcs11:token=Foo;id=%01
   --cert pkcs11:token=Foo;id=%01

tpm2-provider recognizes a custom URI "handle:<hex>":

   --key handle:0x81000000

These examples assume that required providers, if any, are loaded and configured.

Change-Id: I82b32d5ab472926e7889a5f4a90caba14231879a

### DIFF
--- a/src/openvpn/options.c
+++ b/src/openvpn/options.c
@@ -567,10 +567,10 @@ static const char usage_message[] =
     "--dh file       : File containing Diffie Hellman parameters\n"
     "                  in .pem format (for --tls-server only).\n"
     "                  Use \"openssl dhparam -out dh1024.pem 1024\" to generate.\n"
-    "--cert file     : Local certificate in .pem format -- must be signed\n"
-    "                  by a Certificate Authority in --ca file.\n"
+    "--cert file     : Local certificate in .pem format or a URI -- must be signed\n"
+    "                  by a Certificate Authority in --ca file used by the peer.\n"
     "--extra-certs file : one or more PEM certs that complete the cert chain.\n"
-    "--key file      : Local private key in .pem format.\n"
+    "--key file      : Local private key in .pem format or a URI.\n"
     "--tls-version-min <version> ['or-highest'] : sets the minimum TLS version we\n"
     "    will accept from the peer.  If version is unrecognized and 'or-highest'\n"
     "    is specified, require max TLS version supported by SSL implementation.\n"
@@ -3851,6 +3851,7 @@ options_postprocess_mutate(struct options *o, struct env_set *es)
 #define CHKACC_FILEXSTWR (1<<2)  /** If file exists, is it writable? */
 #define CHKACC_ACPTSTDIN (1<<3)  /** If filename is stdin, it's allowed and "exists" */
 #define CHKACC_PRIVATE (1<<4)    /** Warn if this (private) file is group/others accessible */
+#define CHKACC_ACCEPT_URI (1<<5) /** If filename is a URI, no check is done unless it starts with file: */
 
 static bool
 check_file_access(const int type, const char *file, const int mode, const char *opt)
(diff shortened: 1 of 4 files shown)


src/openvpn/ssl_openssl.c
```diff
@@ -959,9 +1162,14 @@ tls_ctx_load_cert_file(struct tls_root_ctx *ctx, const char *cert_file,
...
+        tls_ctx_load_cert_uri(ctx, cert_file);
+        return;
+    }
     else
     {
-        in = BIO_new_file(cert_file, "r");
+        in = BIO_new_file((char *) cert_file, "r");
     }
 
     if (in == NULL)
```
schwabe, line 1175: Maybe move this following part into a separate function like `tls_ctx_load_cert_bio` and then used that afer the `in =` in these two branches that use that? That will make logic here easier to understand and indicate that it is similar to the `tls_ctx_load_cert_uri` function in what it is doing?

src/openvpn/ssl_openssl.c
```diff
@@ -945,6 +1056,98 @@ tls_ctx_add_extra_certs(struct tls_root_ctx *ctx, BIO *bio, bool optional)
...
+    }
+    else
+    {
+        crypto_print_openssl_errors(M_DEBUG);
+    }
+
+    UI_destroy_method(ui_method);
+    OSSL_STORE_INFO_free(info);
+    OSSL_STORE_close(store_ctx);
+#endif /* defined(HAVE_OPENSSL_STORE_API */
```
schwabe, line 1148: maybe 

```
#else
ASSERT(0);
```
?

It doesn't look like if the ifdef is false that this function should ever be called.

schwabe:
Overall I like the approach. I think --crl could also be extended in the same way but I think that would be more an academic excercise since I cannot really came up with a use case for when CRL is not a file or inline. I went through the code and commented on a few things that felt odd or could be better. Also OpenSSL's API are  sometimes wacky but that is kind of expected...

src/openvpn/options.c
```diff
@@ -3851,6 +3851,7 @@ options_postprocess_mutate(struct options *o, struct env_set *es)
 #define CHKACC_FILEXSTWR (1<<2)  /** If file exists, is it writable? */
 #define CHKACC_ACPTSTDIN (1<<3)  /** If filename is stdin, it's allowed and "exists" */
 #define CHKACC_PRIVATE (1<<4)    /** Warn if this (private) file is group/others accessible */
+#define CHKACC_ACPTURI (1<<5)    /** If filename is a URI, no check is done unless it starts with file: */
```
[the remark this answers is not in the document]
selvanair: I was mimicking "CHKACC_ACPTSTDIN" two lines above. Even for me it took a couple of tries to get that right :)
I'm all in for CHKACC_ACCEPT_URI, and will be so in v2.

src/openvpn/ssl_openssl.c
selvanair: I like the idea but we may lose the ability to print the file name in the error message if loading fails. May be I can name it as "tls_ctx_load_cert_pem()" and pass the filename itself instead of bio. Let me see..

selvanair (commented):
Thanks. All good suggestions. Do you want v2 here or push to gerrit -- I don

src/openvpn/ssl_openssl.c
selvanair: Good idea. Instead of passing "in" branched to tls_ctx_load_cert_pem_file() when URI is not supported or if inline. That allows us to continue printing file name and inline flag in the error messages.

flichtenheld:
Merged as 3512e8d3ada4fa7d04925a89fd9f3669655c7887

PR #630, opened 2024-10-22
Pass environment needed for systemd-ask-pass

Firstyear:
systemd-ask-pass allows the pin to be provided by the systemd-credentials mechanism, but that requires the environment variable $CREDENTIALS_DIRECTORY to be set. Openvpn when launched by systemd does have this environment variable available, but it doesn't pass it through to systemd-ask-pass meaning that systemd-ask-pass is unable to load the pin from
systemd-credentials.

This change passes the $CREDENTIALS_DIRECTORY environment variable to the execed systemd-ask-pass invocation so that
systemd-credentials can be used for non-interactive pin entry.

### DIFF
--- a/src/openvpn/console_systemd.c
+++ b/src/openvpn/console_systemd.c
@@ -61,6 +61,9 @@ get_console_input_systemd(const char *prompt, const bool echo, char *input, cons
     int std_out;
     bool ret = false;
     struct argv argv = argv_new();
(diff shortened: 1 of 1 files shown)


Firstyear:
I won't be able to submit this to a mailing list, as I don't have a git email work flow setup. Feel free to send this on my behalf.

PR #634, opened 2024-10-26
Improve shuffling algorithm of connection list

Hurukawa2121:
I propose implementing the Fisher-Yates shuffle algorithm to improve the shuffling of the connection target list.

You can run [this simple simulation](http://tpcg.io/5SPPET) to intuitively understand how much bias is produced by the current method and how much the Fisher-Yates shuffle algorithm improves it.

### Issue:

In the current shuffle algorithm, each element `l->array[i]` is swapped with a random element at index `j` in the entire array.
However, this method does not produce all permutations uniformly, resulting in a bias where some permutations occur more frequently than others.

### Solution:

To ensure all permutations are generated with equal probability, I implemented the Fisher-Yates shuffle algorithm.

#### Intuitive Explanation of Uniformity:

In this algorithm, we start from the end of the array and move towards the beginning.
At each step `i`, we randomly select an index `j` between `0` and `i` (inclusive) and swap `l->array[i]` with `l->array[j]`. This means we're randomly choosing an element from the unshuffled portion and placing it at position `i`.
Since there's only one way to obtain each permutation through these swaps, all permutations occur with equal probability in theory.

#### Proof of Uniformity:

At each iteration `i`, there are exactly `i + 1` choices for `j`, so the probability of selecting any particular `j` is $`\frac{1}{i + 1}`$.
Since the choices at each iteration are independent, the total probability of any specific permutation is:

$`
P = \prod_{k=1}^n \frac{1}{k} = \frac{1}{n!}
`$

Therefore, it is concluded that each permutation occurs with equal probability.

### DIFF
--- a/src/openvpn/init.c
+++ b/src/openvpn/init.c
@@ -478,9 +478,9 @@ init_connection_list(struct context *c)
     if (c->options.remote_random)
     {
         int i;
-        for (i = 0; i < l->len; ++i)
+        for (i = l->len - 1; i > 0; --i)
         {
-            const int j = get_random() % l->len;
+            const int j = get_random() % (i + 1);
             if (i != j)
             {
                 struct connection_entry *tmp;
--- a/src/openvpn/init.c
+++ b/src/openvpn/init.c
@@ -467,7 +467,14 @@ ce_management_query_remote(struct context *c)
 #endif /* ENABLE_MANAGEMENT */
 
 /*
- * Initialize and possibly randomize connection list.
+ * Initialize and randomize the connection list.
+ * 
+ * Applies the Fisher-Yates shuffle algorithm to ensure all permutations are equally probable,
+ * thereby eliminating shuffling bias in the previous method.
+ * 
+ * The algorithm randomly selects an element from the unshuffled portion and places it at position i.
+ * There's only one way to obtain each permutation through these swaps.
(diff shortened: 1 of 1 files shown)


cron2:
This looks interesting.

To progress, please add some comments to the new code explaining what this does ("Shuffle with Fisher-Yates algorithm which has less bias to ....").  Otherwise this creates just code that is not meaningful to the next person looking at it.

Then, send the patch to the openvpn-devel list, or to our gerrit.  We do not merge patches from GH.

thanks ;-)

Hurukawa2121:
Thanks for your feedback :)

I added comments to the new code explaining how the Fisher-Yates shuffle algorithm generates all permutations with equal probability.
Although I couldn't mathematically quantify the bias introduced by the current method, I attached a simple simulation to illustrate the improvement.

Also, I've just sent the patch to the openvpn-devel list as you suggested.

I apologize for mistakenly sending some emails.

schwabe:
Your comments don't seem to be in the files changed tab, did you forget to push your changes?

Hurukawa2121:
I misunderstood his advice and added the comments to the PR description instead of directly within the code.

I have now updated the code with the necessary comments.
I have also resent the email.

Thanks.

Hurukawa2121:
This PR was applied to master branch.
Thank you so much for your help :-)

PR #668, opened 2025-01-14
dco: fix source IP selection

LGA1150:
If the local option is present, DCO should use it as the source IP, or Linux may pick a different IP as the source IP, breaking the connection.

### DIFF
--- a/src/openvpn/dco.c
+++ b/src/openvpn/dco.c
@@ -562,7 +562,7 @@ dco_multi_get_localaddr(struct multi_context *m, struct multi_instance *mi,
         {
             struct sockaddr_in *sock_in4 = (struct sockaddr_in *)local;
 #if defined(HAVE_IN_PKTINFO) && defined(HAVE_IPI_SPEC_DST)
-            sock_in4->sin_addr = actual->pi.in4.ipi_addr;
+            sock_in4->sin_addr = actual->pi.in4.ipi_spec_dst;
 #elif defined(IP_RECVDSTADDR)
             sock_in4->sin_addr = actual->pi.in4;
 #else


schwabe:
Could you add the scenario when this happens with a small config that allows us to reproduce this issue?

LGA1150:
update: manually rebased to master branch

OS/Kernel: OpenWrt 22.03/Linux 5.10.104 (Minimal kernel config to reduce size), musl libc 1.2.2
OpenVPN: 2.6.12
DCO module: 20240712

There are 2 network interfaces wan and lan. Both have static IPs:
```
wan: 192.168.190.105/24
lan: 192.168.8.1/24
```

server.conf:
```ini
local 192.168.190.105
tls-server
allow-compression no
ca /etc/openvpn/pki/ca.crt
cert /etc/openvpn/pki/issued/server.crt
dev tun
dh /etc/openvpn/pki/dh.pem
keepalive 10 60
key /etc/openvpn/pki/private/server.key
server 10.200.200.0 255.255.255.0
topology subnet
tun-mtu 1448
verb 5
```

client.ovpn:
```ini
remote 192.168.190.105
tun-mtu 1448
allow-compression no
client
dev tun
tls-client
# ca, cert, key omitted
```
Client is at `lan` side with IP `192.168.8.2`.
The initial handshake went through, but once the data path is offloaded, DCO transmit path picked the `lan` interface IP `192.168.4.1` as the source IP, breaking the connection. Adding `disable-dco` to server.conf fixes it.

With some debugging I found that `local` option was not passed to DCO via the netlink attr `OVPN_NEW_PEER_ATTR_LOCAL_IP`. , so Linux picked lan interface IP as the source IP.

LGA1150:
Update: found a similar bug when multihome option is enabled. I will push a new patch later.

LGA1150:
sent to the list
- https://sourceforge.net/p/openvpn/mailman/message/59119029/

LGA1150:
It didn't fail immediately at that step. I was looking at "Set job status" and had no idea what it meant.
Anyway I'll fix the code format.

LGA1150:
Merged as https://github.com/OpenVPN/openvpn/commit/e4beaf7a610793f57984ec8006d26f9a695b0c74

PR #672, opened 2025-01-21
Upstream support for AWS-LC

smittals2:
This PR contains minor changes to OpenVPN to improve compatibility with AWS-LC. It also has a README to explain the build process. Opening the PR to gather feedback before I contribute it through gerrit. Thank you!

----Context for changes in ssl_openssl.c around `int cnum` around line 1900----
This change addresses a subtle behavioral difference between AWS-LC and OpenSSL regarding object ownership semantics in SSL_CTX_set_client_CA_list(ctx->ctx, cert_names).

OpenSSL Behavior:

Stores a reference to the provided cert_names stack
cert_names remains valid after SSL_CTX_set_client_CA_list
AWS-LC Behavior:

Creates a copy of the parameter cert_names (which is a stack of type X509_NAME) and converts it to a stack of CRYPTO_BUFFER (how we internally represent X509_NAME, it's an opaque byte string).
Then frees the original passed in cert_names
After SSL_CTX_set_client_CA_list, cert_names no longer points to valid memory
The proposed changes reorder operations to getting the size of the stack before the set operation as opposed to after the set operation. No operations between the setter and stack size check modify cert_names. Therefore, the logical outcome should remain the same - and this would also handle the subtle behavioral difference in AWS-LC.


src/openvpn/crypto_openssl.c
```diff
@@ -1398,7 +1398,7 @@ ssl_tls1_PRF(const uint8_t *seed, int seed_len, const uint8_t *secret,
 
     return ret;
 }
-#elif !defined(LIBRESSL_VERSION_NUMBER) && !defined(ENABLE_CRYPTO_WOLFSSL)
+#elif !defined(LIBRESSL_VERSION_NUMBER) && !defined(ENABLE_CRYPTO_WOLFSSL) && !defined(OPENSSL_IS_AWSLC)
```
schwabe, line 1401: Since you have custom code later in the #elif, I would prefer to just move the #elif defined(OPENSSL_IS_AWSLC) above this statement so we can avoid adding an additional term to this statement.

src/openvpn/ssl_openssl.c
```diff
@@ -1895,9 +1899,11 @@ tls_ctx_load_ca(struct tls_root_ctx *ctx, const char *ca_file,
...
             sk_X509_INFO_pop_free(info_stack, X509_INFO_free);
         }
-
+        
+        int cnum;
         if (tls_server)
         {
+            cnum = sk_X509_NAME_num(cert_names);
             SSL_CTX_set_client_CA_list(ctx->ctx, cert_names);
         }
```
schwabe, line 1908: This change seems weird. With the old code we add the client CA list and then use sk_X509_NAME_num get the number, while with this patch we change the order. I haven't look exactly into what sk_X509_NAME_num does right now since it is one of the annoying function where finding the man page is hard but doesn't this change the logic?

src/openvpn/crypto_openssl.c
smittal3: Great point, made the change!

src/openvpn/ssl_openssl.c
schwabe: Thanks for the explaination.

src/openvpn/ssl_openssl.c
```diff
@@ -1895,9 +1899,11 @@ tls_ctx_load_ca(struct tls_root_ctx *ctx, const char *ca_file,
             }
             sk_X509_INFO_pop_free(info_stack, X509_INFO_free);
         }
-
+        
+        int cnum;
```
schwabe, line 1903: The extra new line here makes our formatting check fail. You can check the "show output on standard output in the github actions to see the full diff."

smittals2: Got it, removed!

src/openvpn/ssl_openssl.c
smittals2: Yep, added

PR #673, opened 2025-01-21
Add AWS-LC to the OpenVPN CI

smittals2:
This PR adds AWS-LC to the github workflows for OpenVPN. Opening the PR to gather feedback and test the workflow. I will eventually contribute these changes through gerrit. Thank you!

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -424,3 +424,61 @@ jobs:
         run: echo 'RUN_SUDO="sudo -E"' >tests/t_server_null.rc
       - name: make check
         run: make -j3 check VERBOSE=1
+
+  aws-lc:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-22.04]
+        ssllib: [ awslc ]
+        build: [ normal, asan ]
+        include:
+          - build: asan
+            cflags: "-fsanitize=address -fno-sanitize-recover=all  -fno-optimize-sibling-calls -fsanitize-address-use-after-scope -fno-omit-frame-pointer -g -O1"
+            ldflags: -fsanitize=address -fno-sanitize-recover=all
+            cc: clang
+          - build: normal
+            cflags: "-O2 -g"
+            ldflags: ""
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -424,3 +424,61 @@ jobs:
         run: echo 'RUN_SUDO="sudo -E"' >tests/t_server_null.rc
       - name: make check
         run: make -j3 check VERBOSE=1
+
+  aws-lc:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-22.04]
```
schwabe, line 432: Any reason to 22.04 instead of 24.04?

smittal3: No particular reason, I was following the other tests in the file. Updated to 24.04.

Is there a way to trigger a CI workflow on this PR to ensure this workflow executes successfully/debug any issues?

schwabe: I can manually trigger it and approved the run but I am not sure why that permission does not stick and I have to do it everytime again.

schwabe: image

It currently shows that there is an error. https://github.com/OpenVPN/openvpn/actions/runs/12919940092/workflow

smittal3: Ah I was missing an indent, sorry about that. Ready to run again!

smittals2: Hmm I'm not sure why the asan build is failing. Threads not found sounds like an OS/CMake issue. 

We test Ubuntu 24.04 with Clang 18 in AWS-LC's CI and that build works. The only difference I noticed is the CXX compiler which I have modified in a new commit. 

Please retry the workflow. If this doesn't work, I will remove the asan dimension.

smittals2:
Hi, somehow an extra space snuck in to the linker flags. I believe this was causing the build failures. Requesting another run of the CI before I submit to mailing list. Thanks!

smittals2:
Submitted the latest version to the mailing list

Issue #683, opened 2025-02-06
DCO + bind-dev, initial connection binds to interface, but switches to default VRF


BCurrell:
**Describe the bug**

When running OpenVPN with DCO enabled, using the bind-dev config directive is not working as expected. The initial connection starts using the interface configured with bind-dev, however after a while, this switches to using the default route interface in the default VRF.

I believe this could be related to when the connection is established, or perhaps if / when the interface offloads to the kernel.

Verified by disabling DCO, when running in TUN mode, the bind-dev config directive works correctly. The connection remains attached to the defined interface.

**To Reproduce**

DCO config, server:
```none
dev-type tun
dev dco01
port 4001
proto udp

data-ciphers AES-256-GCM:AES-128-GCM:?CHACHA20-POLY1305
persist-key
remote-cert-tls client
tls-server

ca /etc/openvpn/keys/rnd/ca.crt
dh /etc/openvpn/dh4096.pem
cert /etc/openvpn/keys/rnd/server.crt
key /etc/openvpn/keys/rnd/server.key

allow-compression no
connect-retry 5 5
fast-io
float
mssfix 1400
max-clients 1
persist-tun
verb 4
```

DCO config, client:
```none
dev-type tun
dev dco01
port 4001
proto udp

# Testing by binding to a VRF
bind-dev TEST_02
# Also tested with the physical interface within the VRF
# bind-dev eth04
remote <endpoint>

data-ciphers AES-256-GCM:AES-128-GCM:?CHACHA20-POLY1305
persist-key
remote-cert-tls server
tls-client

ca /etc/openvpn/keys/rnd/ca.crt
#dh /etc/openvpn/dh4096.pem
cert /etc/openvpn/keys/rnd/client.crt
key /etc/openvpn/keys/rnd/client.key

allow-compression no
connect-retry 5 5
fast-io
float
mssfix 1400
persist-tun
verb 4
```

For the TUN configs, I just added `disable-dco`, the rest was left alone.

In my environment, I use networkd to handle the interface configuration. I can provide the networkd config files with the VRF declaration and interface configuration if required.

**Expected behavior**

The tunnel connects over the bound interface, and continues to use this interface for the whole session.

**Version information (please complete the following information):**
 - OS: Debian 12.9
 - OpenVPN version:
```none
[...]

PR #690, opened 2025-02-16
Add check for bind-dev in DCO options

BCurrell:

### DIFF
--- a/src/openvpn/dco.c
+++ b/src/openvpn/dco.c
@@ -346,6 +346,12 @@ dco_check_startup_option(int msglevel, const struct options *o)
         }
     }
 
+    if (o->bind_dev)
+    {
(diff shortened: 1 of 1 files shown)


BCurrell:
Following from #683, add an option to check for the `bind-dev` option and disable DCO. I'm not sure if this is the correct function to add it to.

Also, isn't `bind-dev` specific to Linux, so do I need to add it to an if block checking for Linux?

cron2:
@ordex I don't think we should disable DCO in this case.  We should fix it inside DCO, not weasel out :-)

ordex:
> @ordex I don't think we should disable DCO in this case. We should fix it inside DCO, not weasel out :-)

that's true, but for the time being it may be easier to let people know that this option is not supported.
Adding support to DCO (ovpn-dco or ovpn? ovpn-dco is in maintenance mode right now) may requires till some time.

PR #772, opened 2025-06-25
Remove unnecessary route delay for Windows 7 and newer.

bernerdad:
This PR considers removing a 5 second delay on Windows when setting up the tunnel.  We have been using this change for 3+ years in the Windscribe desktop app on Windows 10/11, with no adverse impact.

### DIFF
--- a/src/openvpn/options.c
+++ b/src/openvpn/options.c
@@ -3238,18 +3238,6 @@ options_postprocess_mutate_invariant(struct options *options)
         options->tuntap_options.ip_win32_type = IPW32_SET_NETSH;
     }
 
-    if ((dev == DEV_TYPE_TUN || dev == DEV_TYPE_TAP) && !options->route_delay_defined)
-    {
-        /* delay may only be necessary when we perform DHCP handshake */
-        const bool dhcp = (options->tuntap_options.ip_win32_type == IPW32_SET_DHCP_MASQ)
-                          || (options->tuntap_options.ip_win32_type == IPW32_SET_ADAPTIVE);
-        if ((options->mode == MODE_POINT_TO_POINT) && dhcp && (win32_version_info() <= WIN_VISTA))
-        {
-            options->route_delay_defined = true;
-            options->route_delay = 5; /* Vista sometimes has a race without this */
-        }
-    }
-
     if (options->ifconfig_noexec)
     {
         options->tuntap_options.ip_win32_type = IPW32_SET_MANUAL;


schwabe:
Could you please explain how you come to the conclusion that this is no longer necessary and how you tested that?

src/openvpn/options.c
```diff
@@ -3243,7 +3243,7 @@ options_postprocess_mutate_invariant(struct options *options)
         /* delay may only be necessary when we perform DHCP handshake */
         const bool dhcp = (options->tuntap_options.ip_win32_type == IPW32_SET_DHCP_MASQ)
                           || (options->tuntap_options.ip_win32_type == IPW32_SET_ADAPTIVE);
-        if ((options->mode == MODE_POINT_TO_POINT) && dhcp)
+        if ((options->mode == MODE_POINT_TO_POINT) && dhcp && (win32_version_info() <= WIN_VISTA))
```
schwabe, line 3246: we don't support anything older than Vista anymore and probably not even Vista at this point. So you really would need to remove the whole block instead, instead of basically making is dead code.

schwabe:
Yes. But your windscribe tunnel is only one very specific tunnel configuration. So you basically have only tested it in one very specific configuration that removing the delay works. Without really knowing  what the specific problem was that this delay address, it is a hard to judge it if it is safe to remove nowadays.

Also if such a behavioural change is itroduced, it should be mentioned in the Changes.rst

cron2:
> Yes. But your windscribe tunnel is only one very specific tunnel configuration. So you basically have only tested it in one very specific configuration that removing the delay works. Without really knowing what the specific problem was that this delay address, it is a hard to judge it if it is safe to remove nowadays.

The overall point is somewhat moot, as this is in a code path that nobody really gets to use anymore in new OpenVPN installs - it's using TAP+DHCP, while really everything should use win-dco and ipapi (via iservice) these days...  which does not suffer from automatic route-delay settings for DHCP.

Removing code paths that made sense on long-forgotten Windows parts is actually a good thing :-) - having a bit more testing would be good, though.

> Also if such a behavioural change is itroduced, it should be mentioned in the Changes.rst

Indeed.  But since half the regular contributors send stuff without Changes.rst updates these days, it's not something that would hold up the patch if acceptable otherwise.

cron2:
Since nothing has happened here since June, I'll just proceed to close this.

For 2.8, the plan is to remove all interface-/route-handling code from OpenVPN (and fully rely on the interactive service), and also get rid of "fake DHCP server to make tap interfaces happy" mode - which would get rid of the route-delay as well.

PR #802, opened 2025-07-27
Update systemd service name param to match command

Terrance:
The service name displays `%I` which invokes systemd's path mangling (notably, converting `-` to `/`), suggesting you need to provide an encoded parameter (via e.g. `systemd-escape`), but the start command itself uses `%i` which doesn't do the conversion.

This updates the service name to match the start command.

---

Raising here first as I'm not sure what (if anything) else needs doing past modifying the service templates.

### DIFF
--- a/distro/systemd/openvpn-client@.service.in
+++ b/distro/systemd/openvpn-client@.service.in
@@ -1,5 +1,5 @@
 [Unit]
-Description=OpenVPN tunnel for %I
+Description=OpenVPN tunnel for %i
 After=network-online.target
 Wants=network-online.target
 Documentation=man:openvpn(8)
--- a/distro/systemd/openvpn-server@.service.in
+++ b/distro/systemd/openvpn-server@.service.in
@@ -1,5 +1,5 @@
 [Unit]
-Description=OpenVPN service for %I
+Description=OpenVPN service for %i
 After=network-online.target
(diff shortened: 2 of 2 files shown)


flichtenheld:
Might have been better to go for `%I` for both but at this point this would probably break existing installations. So this change makes sense to me.

Terrance:
Indeed, that would be a breaking change for anyone using dashes (and likely other reserved characters) in their configuration file names.

I also think systemd's `%I` leads to surprising behaviour in general -- most template units I've seen don't use it, and having both `%I` and `%i` means a user aware of them has to consult the unit file itself to check which it uses.

Terrance:
I submitted a copy of the patch to the mailing list a few hours ago, but I can't see it in the web archive -- has it been held for moderation, or is the archive just slow to update?

cron2:
> I submitted a copy of the patch to the mailing list a few hours ago, but I can't see it in the web archive -- has it been held for moderation, or is the archive just slow to update?

The list uses greylisting, so for infrequent/first posters, it can take a few hours to show up.  Also, you need to be subscribed to be allowed to post (but if that is the reason, you should see a bounce mail).

Terrance:
Merged as 8854550d4da0b29af122723def898d38e6b3fea2 🙂

PR #809, opened 2025-08-04
Increase systemd service process limit

peterfab9845:
resolvconf ends up calling itself several times due to subscriber scripts and may go over 10 processes, resulting in the below message and ultimately an empty resolv.conf:
```
/bin/resolvconf: fork: retry: Resource temporarily unavailable
```


flichtenheld:
Hmm, I wonder whether it makes sense to increase this value or whether we should just remove it entirely. It is clearly a value that depends very much on the individual setup, especially which user openvpn is running under. 10 is probably a value that was chosen to be "high enough that it doesn't usually cause problems". To be fair, I haven't seen any other reports about this, so in most cases it seems indeed to work fine. (at least in older versions)

peterfab9845:
In my case, openvpn is the only thing running under its user. However, if the decision is to increase rather than remove the limit, I suspect it should be TasksMax=N (limited per service) rather than LimitNPROC=N (per user, as you mentioned).

flichtenheld:
We have #929 now so clearly there is some interest in getting this default changed.

flichtenheld:
I have now proposed https://gerrit.openvpn.net/c/openvpn/+/1539 to fix this issue.

peterfab9845:
I'm not sure that changing from LimitNPROC=10 to TasksMax=10 resolves the original issue I ran into. I'm only running one instance of the client, but I still hit 10 processes.

flichtenheld:
> I'm not sure that changing from LimitNPROC=10 to TasksMax=10 resolves the original issue I ran into. I'm only running one instance of the client, but I still hit 10 processes.

Right.

flichtenheld:
Combined both of the changes, merged to release/2.7 as 8a156abb4b40dee63174fd45fe66e4591409b497

PR #814, opened 2025-08-06
bulk-mode

stoops:
Bulk mode - Feature request

Blog post: https://fossjon.com/2025/08/05/modifying-openvpn-source-code-to-allow-for-bulk-reads-max-mtu-and-jumbo-tcp-for-highly-improved-performance/

### DIFF
--- a/src/openvpn/forward.c
+++ b/src/openvpn/forward.c
@@ -45,6 +45,9 @@
 
 #include "memdbg.h"
 
+#include <sys/select.h>
+#include <sys/time.h>
+
 counter_type link_read_bytes_global;  /* GLOBAL */
 counter_type link_write_bytes_global; /* GLOBAL */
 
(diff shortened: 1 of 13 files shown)


stoops:
Edit/Update: I spent several days and reworked the bulk + mtio mode code a little bit and am now getting almost full line speed that my wifi / internet connection will allow for which is much better/nicer to use as my main network wide TCP VPN. I don't have an official comparison speed test yet as I don't have a wired client available here at home. 
I was able to remap the tun read/write file descriptors on each of the multiple threads and create a an extra management thread which is dedicated to bulk reading and filling up the each of the threaded bulk context buffers and then signalling them all for parallel processing and socket transmission with much larger payload sizes without any thread locks needed. The code seems to be working way better now as this was the first time for me going through the code base and finding out the edge cases here and there! 
Thanks again all for the project work. :)

PR #836, opened 2025-09-11
WIP: Fix auth-user-pass with disabled pull

### LINKED ISSUE #835: 2.6.14 "--auth-user-pass requiring --pull"???
**IMPORTANT NOTE**
Bugs about OpenVPN Access Server, OpenVPN Connect or any other product by OpenVPN Inc. should be directly reported to OpenVPN Inc. at https://support.openvpn.net

**Describe the bug**
Either a config file parsing bug or a documentation issue.

**To Reproduce**
1. Setup a small example deployment with one server and client, fitting the below client config.
2. Try to connect to the server (should work without additional properties, except filling in ca and tls-auth).
3. Now replace `client` and with `tls-client` and `pull`. Docs say this is all that `client` does [Link](https://openvpn.net/community-docs/community-articles/openvpn-2-6-manual.html#:~:text=the%20management%20interface.-,%2D%2Dclient,-A%20helper%20directive).
4. Now add `pull-filter ignore ""` (Config should still work and connect properly). Docs don't mention that anything is except from getting filtered neither at [--pull](https://openvpn.net/community-docs/community-articles/openvpn-2-6-manual.html#:~:text=to%20prevent%20confusion.-,%2D%2Dpull,-This%20option%20must) nor [--pull-filter args](https://openvpn.net/community-docs/community-articles/openvpn-2-6-manual.html#:~:text=client%27s%20routing%20table.-,%2D%2Dpull%2Dfilter%20args,-Filter%20options%20on).
5. Next try to remove the `pull` entirely as the `pull-filter ignore ""` should render it entirely useless anyway. The docs for [--pull](https://openvpn.net/community-docs/community-articles/openvpn-2-6-manual.html#:~:text=to%20prevent%20confusion.-,%2D%2Dpull,-This%20option%20must) even state: 
> so you should not use --pull or --client in situations where you don't trust the server to have control over the client's routing table

But once these two lines are removed parsing the configuration file fails for undocumented reasons with:
> Options error: You must define certificate file (--cert) or PKCS#12 file (--pkcs12)
> Use --help for more information.

Then one may think "just point it towards the CA file or /dev/null". When `--cert /dev/null` is added a similar message but with `--key` instead of `--cert` is shown. Trying to set that one to `/dev/null` as well results in another error message without any documentation (Setting `--pkcs12 /dev/null` instead behaves exactly like `--cert /dev/null --key /dev/null` in this regard):
> Options error: --auth-user-pass requires --pull
> Use --help for more information.

[--auth-user-pass](https://openvpn.net/community-docs/community-articles/openvpn-2-6-manual.html#:~:text=be%20base64%20encoded.-,%2D%2Dauth%2Duser%2Dpass,-Authenticate%20with%20server) within the docs doesn't mention anything about this requirement.

Also even if this is intended for whatever reason the fact that it doesn't work when `pull` is missing but does when `pull-filter ignore ""` definitely needs further elaboration. Same for why cert/key or pkcs12 suddenly becomes required when a client doesn't pull options from the server (esp. when looking at what options get logged as having been sent by the server. See below)

6. Out of desperation try to add all of the options the server (supposedly according to the logs) sent directly into the config file => Same parsing error as above.

Working client config before above steps to start with:
```
client

dev tun
proto udp
remote example.invalid 1194
auth-user-pass 
link-mtu 1500
mtu-test
resolv-retry infinite
nobind
user nobody
group nogroup
persist-key
persist-tun
key-direction 1

<ca>
-----BEGIN CERTIFICATE-----
(...)
-----END CERTIFICATE-----
</ca>

remote-cert-tls server

<tls-auth>
# 2048 bit OpenVPN static key
#
-----BEGIN OpenVPN Static key V1-----
(...)
-----END OpenVPN Static key V1-----
</tls-auth>

data-ciphers AES-256-CBC
data-ciphers-fallback AES-256-CBC
verb 5
dhcp-option DOMAIN-ROUTE .
```

Even though the mentioned error is while parsing the client configuration before any connection is attempted here are the options the server would be pushing (as logged by the pull-filter option):
```
# What the server would push:
#route 10.0.0.0 255.255.0.0
#route 192.168.0.0 255.255.255.0
#dhcp-option DNS 10.0.0.11
#dhcp-option DNS 10.0.0.12
#route-gateway 172.16.0.1
#topology subnet
#ping 10
#ping-restart 120
#ifconfig 172.16.0.20 255.255.255.0
#peer-id 4
#cipher AES-256-CBC
```

**Expected behavior**
Either documentation stating what side effects or hidden configuration options get set/unset when comparing:
a) `pull` being omitted with `pull` and `pull-filter ignore ""` (what is different, why, and where to find it in the logs?)
b) The configuration with all of the pulled (and logged) options being included in the config (instead of using `pull`) with one that only contains `pull`

The expected behaviour would be consistent behaviour. As both in a should behave exactly the same as should the two in b.

And/Or documentation needs to be expanded to state that what's different and why.

Also in case there is some technical reason for why `pull` is demanded here, a less intrusive option should be provided that *only* gets the minimal required properties (or add the `pull-filter ignore ""`-trick to the docs as the "official" solution for these situations together with an explanation what it modifies).

**Version information (please complete the following information):**
 - OS: MacOS 15.6.1
 - OpenVPN version: OpenVPN 2.6.14 x86_64-apple-darwin [SSL (OpenSSL)] [LZO] [LZ4] [MH/RECVDA] [AEAD]
 - library versions: OpenSSL 3.0.16 11 Feb 2025, LZO 2.10

**Additional context**
I'm connecting to a openvpn server of a 3rd party company to access some specific systems only. I do not want them to fuck up my network configuration or being considered authoritative or trusted for pushing ANY configuration including e.g. DNS or full-tunnel (esp. because some of the routes they push overlay with other networks I need to access). I'll manually configure the routes I want to have and everything else is undesired here too.

agowa:
Rework options parsing for auth-user-pass to work without --pull.

Fixes #835

### DIFF
--- a/src/openvpn/options.c
+++ b/src/openvpn/options.c
@@ -2763,6 +2763,7 @@ options_postprocess_verify_ce(const struct options *options, const struct connec
 #ifdef ENABLE_CRYPTOAPI
             MUST_BE_UNDEF(cryptoapi_cert, "cryptoapicert");
 #endif
+            MUST_BE_UNDEF(auth_user_pass_file, "auth_user_pass");
         }
         else
 #endif /* ifdef ENABLE_PKCS11 */
@@ -2776,6 +2777,7 @@ options_postprocess_verify_ce(const struct options *options, const struct connec
             MUST_BE_UNDEF(pkcs12_file, "pkcs12");
             MUST_BE_FALSE(options->management_flags & MF_EXTERNAL_KEY, "management-external-key");
             MUST_BE_FALSE(options->management_flags & MF_EXTERNAL_CERT, "management-external-cert");
+            MUST_BE_UNDEF(auth_user_pass_file, "auth_user_pass");
         }
         else
 #endif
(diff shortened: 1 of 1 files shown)


cron2:
The diff looks much larger than it needs to be, and many unrelated changes (like the last hunk about "ncp-fallback") - seems like a git rebase accident?

agowa:
Yea probably. But when you look into the main ticket it appears this change isn't even wanted. People appear to like giving random servers they connect to authority to push random configurations to their devices for whatever reason. I'd call that an anti-feature and in a case like here where it can't even be disabled a (to be fair quite limited) backdoor...

cron2:
> Yea probably. But when you look into the main ticket it appears this change isn't even wanted. People appear to like giving random servers they connect to authority to push random configurations to their devices for whatever reason. I'd call that an anti-feature and in a case like here where it can't even be disabled a (to be fair quite limited) backdoor...

`--pull-filter` exists and does exactly what you want: filter out specific or all configs pushed by the server.

This said, if you want to contribute code, it needs to be correct and not contain totally unrelated changes - this is something I can not work with.

schwabe:
Also I still don't think enabling auth-user-pass is the right approach if you want more control in a situation where the PUSH_REQUEST/PUSH_REPLY protocol is used. Yes this might work in your special setup but generally a client in p2p does not work well against a p2mp server since a lot of the negotiation will not work.

@agowa for your scenario that the `route-nopull` should work. If not, please describe what you are missing in that feature.

PR #895, opened 2025-11-04
Invalid Index - Bug Fix

stoops:
Observed that there is an edge case where if the session ID is not found (i==3) && it is not a hard reset then the code path will then enter into the else block with (i==TM_SIZE==3) and enter into the multi->session[i] array (session[TM_SIZE]) which is an invalid index access!

### DIFF
--- a/src/openvpn/ssl.c
+++ b/src/openvpn/ssl.c
@@ -3679,7 +3679,7 @@ tls_pre_decrypt(struct tls_multi *multi, const struct link_socket_actual *from, 
      * multi->session: Possible initial packet. New sessions always start
      * as TM_INITIAL
      */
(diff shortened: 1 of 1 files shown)


schwabe:
This is not the correct fix as it changes the semantics. I have already a correct fix prepared that should be push to the OpenVPN repo in the next few days.

stoops:
Hi Arne, thanks for taking a look at it. Yeah this was just my temporary workaround fix to help protect the next else block but I wasn't sure if there were any edge cases or code paths that might not be covered by changing the conditionals. I can close this PR, thanks again!

cron2:
https://gerrit.openvpn.net/c/openvpn/+/1373

Credits to @stoops (and ZeroPath, found twice in very short time) will be added to the commit message on merge

PR #925, opened 2025-12-01
PUSH_UPDATE: fix option reset logic in continuation messages

maurice2k:
Previously, the logic for resetting push options (like 'route') was based on `update_options_found` which was local to `apply_push_options`. This meant that if a PUSH_UPDATE was split across multiple continuation messages, the state was lost, causing routes to be reset multiple times (once per message chunk) rather than once per update sequence.

This patch moves the state tracking to `struct options` as `push_update_options_found`, allowing it to persist across the entire PUSH_UPDATE sequence.

This fixes an issue where large route lists sent via PUSH_UPDATE would result in only the last chunk's routes being applied, or previous routes being continuously deleted and re-added.

Added unit test `test_incoming_push_continuation_route_accumulation` to verify the fix.

### DIFF
--- a/src/openvpn/options.c
+++ b/src/openvpn/options.c
@@ -3234,6 +3234,7 @@ pre_connect_restore(struct options *o, struct gc_arena *gc)
 
     o->push_continuation = 0;
     o->push_option_types_found = 0;
+    o->push_update_options_found = 0;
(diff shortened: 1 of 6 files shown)


maurice2k:
https://sourceforge.net/p/openvpn/mailman/message/59267208/

PR #930, opened 2025-12-06
Prevent crash on invalid server-ipv6 argument

klemensn:
`get_addr_generic()` expects `openvpn_getaddrinfo()` to return a newly
allocated struct, but getaddrinfo(3) failure leaves `*ai = NULL` as-is.

Unlike free(3), freegetaddrinfo(3) requires a valid struct, thus the
following is enough to trigger a NULL pointer dereference in libc:

```
$ openvpn --server-ipv6 ''
2025-12-06 11:59:18 RESOLVE: Cannot resolve host address: :[AF_INET6] (no address associated with name)
Segmentation fault (core dumped)
```

Guard against empty `ai`, i.e. failure, like similar code already does:

```
$ ./openvpn --server-ipv6 ''
2025-12-06 12:05:11 RESOLVE: Cannot resolve host address: :[AF_INET6] (no address associated with name)
Options error: error parsing --server-ipv6 parameter
Use --help for more information.
```

Spotted through a configuration typo "server-ipv6 fd00:/64" with 2.6.17,
reproduced with and tested against 2.7rc3 on OpenBSD/amd64 7.8-current.

Signed-off-by: Klemens Nanni <kn@openbsd.org>

### DIFF
--- a/src/openvpn/socket.c
+++ b/src/openvpn/socket.c
@@ -189,7 +189,10 @@ get_addr_generic(sa_family_t af, unsigned int flags, const char *hostname, void 
         *sep = '/';
     }
 out:
-    freeaddrinfo(ai);
+    if (ai)
+    {
+        freeaddrinfo(ai);
+    }
     free(var_host);
 
     return ret;


cron2:
Interesting find.

This looks a bit like an OpenBSD stubbornness...  since it didn't crash running the same test on Linux or FreeBSD, I decided to have a look into the library implementations to better understand why.

FreeBSD 14.3 source says

```
/*
 * Issues to be discussed:
...
 * - freeaddrinfo(NULL).  RFC2553 is silent about it.  XNET 5.2 says it is
 *   invalid.  Current code accepts NULL to be compatible with other OSes.
...
void
freeaddrinfo(struct addrinfo *ai)
{
        struct addrinfo *next;

        while (ai != NULL) {
                next = ai->ai_next;
                free(ai->ai_canonname);
                /* no need to free(ai->ai_addr) */
                free(ai);
                ai = next;
        }
}
```

so NULL is actually safe with this implementation.  Still, given that there is standards ambiguity, it makes sense to avoid that question on our side.

(I have no extracted OpenBSD source at hand to see how they do it)

Please send to gerrit or to the openvpn-devel@ list ;-)

cron2:
Yeah, bug indeed, at least inconsistency inside OpenVPN - I quickly checked all other calls to `freeaddrinfo()`, and as far as I could see these are either guarded by `if (ai)` or in a branch that accesses `ai` beforehand, so we known it's not NULL.

klemensn:
> Please send to gerrit or to the openvpn-devel@ list ;-)

Moving it to Gerrit now, thanks.

https://gerrit.openvpn.net/c/openvpn/+/1418

PR #953, opened 2026-01-07
Add --sni option for TLS Server Name Indication support

steffen-heil-secforge:
Add support for sending SNI (Server Name Indication) during TLS handshake. The new --sni option accepts three modes:
- Not set: SNI disabled (default, backwards compatible)
- "auto": Use hostname from --remote directive
- <hostname>: Use specified hostname

Implementation covers both OpenSSL and mbedTLS backends.

### DIFF
--- a/doc/man-sections/tls-options.rst
+++ b/doc/man-sections/tls-options.rst
@@ -408,6 +408,28 @@ certificates and keys: https://github.com/OpenVPN/easy-rsa
     but also the key exchange in TLS 1.3 and using this option improperly
     will disable TLS 1.3.
 
+--sni hostname
+    Set the TLS Server Name Indication (SNI) for the control channel TLS
+    handshake.
+
+    Valid syntax:
+    ::
+
+       sni auto
+       sni hostname
+
+    When set to :code:`auto`, the hostname from the ``--remote`` option is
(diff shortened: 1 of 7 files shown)


schwabe:
Could you explain what the use case for this option is? You have only implemented client side, so it does not offer the typical SNI advantages.

cron2:
Indeed this would need a very strong and compelling use case.

Adding crypto related options is easy, maintaining them for the rest of OpenVPN's life (testing, ensuring nothing breaks on upgrades, etc.) is a burden on us, so it needs to bring a clear benefit for sufficiently-many users.

steffen-heil-secforge:
Sorry, this was an accident, I wanted to create a PR in my own fork before testing. It went here and I did not notice.
Second only when testing now, I realized OpenVPN wraps TLS handshake in custom packets, which really make this change useless.

My intention was to route OpenVPN traffic through nginx using the stream plugin by SNI....
I will close this PR now.

schwabe:
@steffen-heil-secforge no problem. If there is a use case and something that we can actually test/use like haproxy with different OpenVPN server behind it, this is can be accepted. But currently it is hard to test/use this feature.

PR #998, opened 2026-03-06
Initial version of macOS TAP support using feth interfaces

kabassanov:
Initial version of macOS TAP support using feth interfaces

This initial version supports only manually created feth interfaces, called fethX and feth(X+1000). They must be created before openvpn start.

1.
ifconfig feth0 create
ifconfig feth1000 create
ifconfig feth0 peer feth1000
ifconfig feth0 up
ifconfig feth1000 up

2. 
config.ovpn file must contain:
dev feth0
dev-type tap
(eventually) route-delay 10

3. (could be useful)

/usr/sbin/ipconfig set feth0 DHCP
/usr/sbin/ipconfig set feth0 AUTOMATIC-V6

There are still several things to check/improuve:
- dynamic creation of the interfaces within openvpn
- validate corectness of conditional compilation flags (ifdef, if define) added/modified near new code

### DIFF
--- a/src/openvpn/forward.c
+++ b/src/openvpn/forward.c
@@ -2364,10 +2364,12 @@ io_wait(struct context *c, const unsigned int flags)
                 c->c2.event_set_status = ES_TIMEOUT;
             }
         }
+#if !defined(TARGET_DARWIN)
         else
         {
             c->c2.event_set_status = SOCKET_READ;
         }
+#endif
     }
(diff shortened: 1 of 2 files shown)


kabassanov:
Wrong files uploaded... Please, wait some minutes...

kabassanov:
It is OK now. Sorry.

PR #1004, opened 2026-03-23
Use memcmp_constant_time for SHA-256 hash comparisons

kodareef5:
Three security-critical comparisons use `memcmp()` on SHA-256 digests where `memcmp_constant_time()` should be used:

- `ssl_verify.c:244` — certificate fingerprint comparison
- `ssl_mbedtls.c:1038` — TLS certificate hash comparison during renegotiation
- `init.c:2167-2168` — configuration digest comparison

The HMAC verification at `crypto.c:652` already uses `memcmp_constant_time` (CVE-2025-13086 fix). These three sites have the same pattern but were not updated.

### DIFF
--- a/src/openvpn/init.c
+++ b/src/openvpn/init.c
@@ -31,6 +31,7 @@
 #endif
 
 #include "win32.h"
+#include "crypto.h"
 #include "init.h"
 #include "run_command.h"
 #include "sig.h"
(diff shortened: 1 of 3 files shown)


kodareef5:
Thanks for the thorough review, @schwabe. Your points are well taken:

- The tun device options check is a functional comparison, not a security gate — constant-time adds no value there.
- The mbedTLS path is deprecated in newer versions, so hardening it is moot.
- The certificate fingerprint comparison requires a compromised CA to even begin timing analysis, which is already game over.

I was applying the pattern too broadly without considering the actual threat model at each call site. Closing this — appreciate the detailed feedback.

PR #1005, opened 2026-03-23
Clamp epoch key generation count to array bounds

kodareef5:
`epoch_generate_future_receive_keys()` in `src/openvpn/crypto_epoch.c` computes `num_keys_generate` as the difference between desired and current highest epoch keys. A large epoch jump from the peer (e.g., from epoch 100 to 65000) makes `num_keys_generate` much larger than the `epoch_data_keys_future` array size.

The `ASSERT` at line 253 catches this in debug builds but is compiled out in release, leading to out-of-bounds array access in the loop and subsequent `memmove`.

Example:
- `current_decrypt_epoch = 65000` (from peer), `future_count = 16`
- `desired_highest_key = 65016`, `current_highest_key = 116`
- `num_keys_generate = 64900` — far exceeds array size of 16

Fix: clamp `num_keys_generate` to `epoch_data_keys_future_count` before the loop.

### DIFF
--- a/src/openvpn/crypto_epoch.c
+++ b/src/openvpn/crypto_epoch.c
@@ -245,6 +245,12 @@ epoch_generate_future_receive_keys(struct crypto_options *co)
     uint16_t desired_highest_key = current_decrypt_epoch + co->epoch_data_keys_future_count;
     uint16_t num_keys_generate = desired_highest_key - current_highest_key;
 
+    /* Clamp to array bounds to prevent OOB access from large epoch jumps */
+    if (num_keys_generate > co->epoch_data_keys_future_count)
+    {
+        num_keys_generate = co->epoch_data_keys_future_count;
+    }
+
     /* Move the old keys out of the way so the order of keys stays strictly
      * monotonic and consecutive. */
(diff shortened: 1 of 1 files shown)


cron2:
Please do not use KI to submit PRs.

Our `ASSERT()`s never get compiled out.  See commit 88f8edbf7.

kodareef5:
You're right — I missed that OpenVPN's ASSERT is always compiled in, not conditional on NDEBUG like the standard C assert. The existing protection is sufficient. Apologies for the noise, closing this.

schwabe:
> A large epoch jump from the peer (e.g., from epoch 100 to 65000)

This cannot happen as we only jump to keys that are valid. So with default you can only jump from 100 to 116. 

And this statement:

        /* If we have reached the edge of the valid keys we do not return
         * the key anymore since regenerating the new keys would move us
         * over the window of valid keys and would need all kind of
         * special casing, so we stop returning the key in this case */
        if (epoch > (UINT16_MAX - opt->epoch_data_keys_future_count - 1))

Takes care of not accepting any key as valid that would bring the new key generation over the limit.

Issue #1037, opened 2026-05-19
client-nat corrupts UDP checksum when input checksum is 0 (RFC 768 violation)


laviniuc:
**Summary**

client_nat_transform() in src/openvpn/clinat.c unconditionally applies an incremental checksum adjustment to the UDP header checksum field when rewriting IP addresses via client-nat. When the inbound UDP packet has checksum=0 (meaning "no checksum computed" per RFC 768), the incremental adjustment writes a bogus non-zero value, causing the packet to be rejected by strict receiving stacks (notably the Windows TCP/IP stack).

**Affected versions**

Confirmed on 2.5.11 and 2.7.4. Source inspection of master shows the bug is still present.

**RFC reference**

Per RFC 768, a UDP checksum field of 0x0000 indicates "checksum not computed" and must not be treated as a checksum value subject to update. NAT implementations rewriting IP addresses must either leave checksum=0 unchanged or fully recompute the checksum from scratch.

**Reproduction**

Client config includes a client-nat dnat rule. A UDP datagram arrives at the client with udp.check = 0x0000 and a destination address that matches the NAT rule. After client_nat_transform() runs, the UDP checksum is a non-zero, invalid value. The receiving OS drops the packet at UDP checksum validation.

Captures of the same datagram before and after the client:

Before (inner-tunnel capture pre-NAT):

```
45 00 00 2b 3c 4f 00 00 3f 11 7f 59 ac 10 03 00

0a 08 07 02 70 70 e8 90 00 17 00 00 ...
```

                              ^^^^^ udp.check = 0x0000

After (post-NAT, as delivered to host stack):

```
45 00 00 2b 3c 4f 00 00 3e 11 12 36 0a 34 13 00

0a 08 07 02 70 70 e8 90 00 17 91 dc ...
```

                              ^^^^^ udp.check = 0x91dc (corrupted)

Only the destination IP and IP checksum legitimately changed. UDP payload is unchanged. UDP checksum should have remained 0x0000.

Source location

src/openvpn/clinat.c, client_nat_transform(), UDP branch of the protocol-specific checksum adjustment:

else if (h->ip.protocol == OPENVPN_IPPROTO_UDP)

```
{

    if (BLEN(ipbuf) >= sizeof(struct openvpn_iphdr) + sizeof(struct openvpn_udphdr))

    {

        ADJUST_CHECKSUM(accumulate, h->u.udp.check);

    }

}
```

[...]

PR #1038, opened 2026-05-19
Fix UDP checksum handling in clinat.c

laviniuc:
#1037

### DIFF
--- a/src/openvpn/clinat.c
+++ b/src/openvpn/clinat.c
@@ -258,7 +258,22 @@ client_nat_transform(const struct client_nat_option_list *list, struct buffer *i
         {
             if (BLENZ(ipbuf) >= sizeof(struct openvpn_iphdr) + sizeof(struct openvpn_udphdr))
             {
-                ADJUST_CHECKSUM(accumulate, h->u.udp.check);
+                /* RFC 768: a UDP checksum of 0 means "no checksum computed".
+                 * Do not run the incremental adjustment over a non-checksum,
+                 * or we will write a bogus non-zero value into the field. */
+        
+                if (h->u.udp.check != 0)
+                {
+                    ADJUST_CHECKSUM(accumulate, h->u.udp.check);
+        
+                    /* RFC 768: a computed checksum of 0 must be transmitted
+                     * as 0xFFFF, because 0 is reserved to mean "no checksum". */
+        
+                    if (h->u.udp.check == 0)
+                    {
+                        h->u.udp.check = 0xFFFF;
+                    }
+                }
             }
         }
     }


laviniuc:
patch for #1037

src/openvpn/clinat.c
```diff
@@ -258,7 +258,22 @@ client_nat_transform(const struct client_nat_option_list *list, struct buffer *i
...
+                if (h->u.udp.check != 0)
+                {
+                    ADJUST_CHECKSUM(accumulate, h->u.udp.check);
+        
+                    /* RFC 768: a computed checksum of 0 must be transmitted
+                     * as 0xFFFF, because 0 is reserved to mean "no checksum". */
+        
+                    if (h->u.udp.check == 0)
+                    {
+                        h->u.udp.check = 0xFFFF;
```
ordex, line 274: You quoted the RFC saying: "NAT implementations rewriting IP addresses must either leave checksum=0 unchanged or fully recompute the checksum from scratch."

Why setting it to 0xFFFF?

ordex:
I added a comment to your patch. However, please ensure to write a proper commit message and add a Signed-off-by line, as explained in our `Contributing` guide, also the patch should be sent to the openvpn-devel for final review and approval (as mentioned in the template that you re-posted above :))

src/openvpn/clinat.c
laviniuc: that's the check after it's been re-adjusted / recomputed. if it happens to be 0 it should be sent as FFFF

ordex: do you have a reference about this rule that we could add as a comment here?
I was looking up the RFC, but I couldn't find anything about checksum=0.

laviniuc: https://www.rfc-editor.org/rfc/rfc768.html page 2 
and explained here https://people.na.infn.it/~garufi/didattica/CorsoAcq/Trasp/Lezione9/tcpip_ill/udp_user.htm section 11.3

ordex: Yeah, my bad. For whatever reason I mixed up your patch with another change and got off-track.

ordex:
Since you don't want to send the patch to the mailing list, may I know how you wanna be credited (name + email) so I can prepare and send a patch for you?

Thanks!

laviniuc:
> Since you don't want to send the patch to the mailing list, may I know how you wanna be credited (name + email) so I can prepare and send a patch for you?
> 
> Thanks!

thank you,

it is fine by me to leave it annonymous.

ordex:
> it is fine by me to leave it annonymous.

ok, thank you for the fix!

I'll have to sign the patch in any case, because it's required in order to accept the license and get the code merged.

I'll close this issue once the patch is in.

ordex:
patch is on gerrit: https://gerrit.openvpn.net/c/openvpn/+/1681

PR #1044, opened 2026-05-27
proto: correct 802.1Q length check in is_ipv_X

rootvector2:
In TAP mode is_ipv_X() checks an 802.1Q frame against the untagged Ethernet header size but then advances past the larger 18-byte 802.1Q header. A tagged frame of 34 to 37 bytes passes the check yet leaves fewer than sizeof(struct openvpn_iphdr) bytes at the new buffer head, so callers such as client_nat_transform that trust the return value access bytes past the IP header. Use sizeof(struct openvpn_8021qhdr) for the check instead.

### DIFF
--- a/src/openvpn/proto.c
+++ b/src/openvpn/proto.c
@@ -70,7 +70,7 @@ is_ipv_X(int tunnel_type, struct buffer *buf, int ip_ver)
         if (proto == htons(OPENVPN_ETH_P_8021Q))
         {
             const struct openvpn_8021qhdr *evh;
-            if (BLENZ(buf) < sizeof(struct openvpn_ethhdr) + sizeof(struct openvpn_iphdr))
+            if (BLENZ(buf) < sizeof(struct openvpn_8021qhdr) + sizeof(struct openvpn_iphdr))
             {
                 return false;
             }


rootvector2:
any update?

cron2:
Thanks for the finding.  As the welcome message tells you, we don't deal with github PRs, but ask for patches to be sent to our gerrit or to the mailing list.  I can forward to gerrit, but we had lots of other work to do, so this extra step can lead to patches getting overlooked (which is a shame, but days are short).

Before proceeding, I need a `signed-off-by: name <e-mail>` address in the git commit message.  I can add this for you, @rootvector2, but need to know what to put there.

cron2:
Patch went to gerrit, has been reviewed and approved, and is now undergoing final github actions checks.  Expect it to show up in the "real" openvpn repo tomorrow.

cron2:
commit f8b15dad1258d6cf71ffa16f67efd9f8b57b9727 (master)
commit 47c9267120d3ee556c30616e1e0e218b78144441 (release/2.7)
commit 66c7d4352c10ca51c20d1fc34c978164e93d21d4 (release/2.6)
Author: rootvector2
Date:   Wed Jul 15 22:22:02 2026 +0200

     proto: correct 802.1Q length check in is_ipv_X

     Signed-off-by: rootvector2 <dxbnaveed.k@gmail.com>

PR #1050, opened 2026-06-11
ci: add read-only workflow token permissions

Alb3e3:
## Summary

All four GitHub Actions workflows lacked an explicit workflow-level `permissions:` block. Without one, each run inherits the repository default GITHUB_TOKEN permissions, which can be broader than these jobs need.

This revision keeps the patch focused on workflow token permissions and sets the default token scope to read-only repository contents access.

### Changes

- `build.yaml`: add `permissions: contents: read`
- `coverity-scan.yml`: add `permissions: contents: read`
- `doxygen.yml`: add `permissions: contents: read`; the deploy job keeps its existing job-level `pages: write` and `id-token: write` permissions
- `test-ssllib.yml`: add `permissions: contents: read`

### Verification

```bash
python3 - <<'PY'
import pathlib, yaml
for p in pathlib.Path('.github/workflows').glob('*.yml'):
    yaml.safe_load(p.read_text())
for p in pathlib.Path('.github/workflows').glob('*.yaml'):
    yaml.safe_load(p.read_text())
print('yaml ok')
PY
```

Result: `yaml ok`.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -4,6 +4,9 @@ on:
   push:
   pull_request:
 
+permissions:
+  contents: read
+
 jobs:
   clang-format:
     name: Check code style with clang-format
--- a/.github/workflows/test-ssllib.yml
+++ b/.github/workflows/test-ssllib.yml
@@ -1,5 +1,8 @@
 name: test_ssllib
 
+permissions:
+  contents: read
+
 on:
   workflow_call:
     inputs:
--- a/.github/workflows/coverity-scan.yml
+++ b/.github/workflows/coverity-scan.yml
@@ -4,6 +4,9 @@ on:
     - cron: '0 20 * * *' # Daily at 20:00 UTC
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   latest:
     # Running coverity requires the secrets.COVERITY_SCAN_TOKEN token
--- a/.github/workflows/doxygen.yml
+++ b/.github/workflows/doxygen.yml
@@ -3,6 +3,9 @@ on:
   push:
     branches: ["master"]
   workflow_dispatch:
+permissions:
+  contents: read
+
 concurrency:
   group: "pages"
   cancel-in-progress: false


.github/workflows/build.yaml
```diff
@@ -4,6 +4,8 @@ on:
   push:
   pull_request:
 
+permissions: read-all
```
flichtenheld: If we change this I would prefer we document the actually needed permissions.

.github/workflows/test-ssllib.yml
```diff
@@ -65,6 +67,13 @@ jobs:
       CXX: ${{matrix.cxx}}
       UBSAN_OPTIONS: print_stacktrace=1
       LIBPREFIX: /opt/${{inputs.libname}}
+      LIB_CONFIGURE: ${{inputs.libconfigure}}
```
flichtenheld: I don't exactly understand what this protects against? Injection by whom? This is all hard-coded stuff.

flichtenheld (changes requested):

.github/workflows/test-ssllib.yml
```diff
@@ -103,14 +112,14 @@ jobs:
         run: autoreconf -fvi
       - name: "OpenVPN: configure"
         run: |
-          ${{inputs.ovpnconfigureenv}} \
-          ./configure --with-crypto-library=${{inputs.ovpnlibname}} ${{inputs.ovpnconfigureflags}}
+          env -S "$OVPN_CONFIGURE_ENV" \
```
flichtenheld: Does not actually work. See GHA run.

    env: only ${VARNAME} expansion is supported, error at: $LIBPREFIX/include" OPENSSL_LIBS="-L$LIBPREFIX/lib -lssl -lcrypto" LDFLAGS="-Wl,-rpath=$LIBPREFIX/lib"

Alb3e3:
Thanks for the review! I've narrowed this PR to permissions-only and addressed your points:

- **test-ssllib.yml** — dropped the `env`/shell change (the `env:`-only `${VARNAME}` expansion that broke the run) and the hard-coded-input hardening, since as you noted those inputs are all hard-coded. Only the `permissions: { contents: read }` addition remains.
- **Documenting the actually needed permissions** — the default `GITHUB_TOKEN` scope these workflows require is just `contents: read` (checkout only); none of build.yaml, coverity-scan.yml, or test-ssllib.yml write to the repository. The one job that needs more — the `deploy` job in doxygen.yml — already declares its own explicit `permissions: { pages: write, id-token: write }`, and a top-level default does not override a job that sets its own block, so Pages deployment is unaffected.

All CI is green on the current head (342051e). Could you take another look when you have a moment?

PR #1051, opened 2026-06-11
push: guard AUTH_PENDING timeout calculation against unsigned underflow

Alb3e3:
## Problem

`send_auth_pending_messages()` calculates the remaining budget for the
AUTH_PENDING timeout by subtracting the elapsed key-state age from a
maximum derived from the renegotiation and handshake-window settings:

```c
unsigned int max_timeout =
    max_uint(tls_multi->opt.renegotiate_seconds / 2, tls_multi->opt.handshake_window);
max_timeout = max_timeout - (now - ks->initial);   // unsigned underflow here
timeout = min_uint(max_timeout, timeout);
```

Both operands are unsigned.  If the elapsed time exceeds `max_timeout`
(e.g. because a slow authentication back-end has already consumed the
entire window, or the clocks have shifted), the subtraction wraps around
to a value near `UINT_MAX`.  The subsequent `min_uint()` call then has no
effect, and the `AUTH_PENDING,timeout X` message sent to the peer carries
a timeout that is orders of magnitude larger than the intended security
window.  The peer may extend its handshake timeout accordingly, keeping
the connection in a pending-authentication state far beyond the configured
limit.

(The enclosing `#pragma GCC diagnostic ignored "-Wconversion"` block was
suppressing the compiler warning that would otherwise have flagged the
mixed signed/unsigned arithmetic here.)

## Fix

Compute the elapsed duration as a named `unsigned int` and use a
saturating subtraction: when `elapsed >= max_timeout` clamp `max_timeout`
to zero rather than wrapping it.

```c
unsigned int elapsed = (unsigned int)(now - ks->initial);
max_timeout = (elapsed < max_timeout) ? (max_timeout - elapsed) : 0;
timeout = min_uint(max_timeout, timeout);
```

An already-expired window is now correctly reported as zero remaining
time, and the `AUTH_PENDING` timeout sent to the peer is clamped
accordingly.

### DIFF
--- a/src/openvpn/push.c
+++ b/src/openvpn/push.c
@@ -446,10 +446,14 @@ send_auth_pending_messages(struct tls_multi *tls_multi, struct tls_session *sess
     unsigned int proto = extract_iv_proto(peer_info);
 
 
-    /* Calculate the maximum timeout and subtract the time we already waited */
+    /* Calculate the maximum timeout and subtract the time we already waited.
+     * Guard against unsigned underflow: if more time has already elapsed
+     * than max_timeout allows, clamp to zero so the caller's timeout is
+     * also reduced to zero rather than wrapping to UINT_MAX. */
     unsigned int max_timeout =
         max_uint(tls_multi->opt.renegotiate_seconds / 2, tls_multi->opt.handshake_window);
-    max_timeout = max_timeout - (now - ks->initial);
+    unsigned int elapsed = (unsigned int)(now - ks->initial);
+    max_timeout = (elapsed < max_timeout) ? (max_timeout - elapsed) : 0;
     timeout = min_uint(max_timeout, timeout);
 
     struct gc_arena gc = gc_new();


flichtenheld:
FWIW, there is already a proposed patch for this at https://gerrit.openvpn.net/c/openvpn/+/1293

cron2:
Gerrit 1293 has been merged as

commit 8d23076babece34581ae9396b6280eb34c86c422 (HEAD -> master, vweb1/master)
Author: Frank Lichtenheld <frank@lichtenheld.com>
Date:   Fri Jun 12 13:33:02 2026 +0200

    push: Fix conversion issues related to timeout in send_auth_pending_messages
    
    Add additional checking to make sure that the required
    casts are safe.

but it touches a bit more than just the push.c / `send_auth_pending_messages()` timeout underflow.

Shall we backport just that change to 2.7 (and 2.6)?

Alb3e3:
Apologies for the slow reply.

Confirming the current state, checked against the branch tips as of 2026-07-31:

- `master` has the fix from Gerrit 1293 (`8d23076`) — `send_auth_pending_messages()` now computes `time_elapsed` as a signed `time_t` and returns early when it is negative or `>= max_timeout`.
- `release/2.7` and `release/2.6` both still have the original line:

  ```c
  max_timeout = max_timeout - (now - ks->initial);
  ```

So yes, the backport is still needed on both, and the answer to "just that change" is that the `send_auth_pending_messages()` hunk is self-contained — it does not depend on the rest of `8d23076`, which touches unrelated conversions.

Since `master` is already fixed, this PR is obsolete as it stands (and now conflicts), so I'm closing it. If a backport to 2.7/2.6 would be useful from me rather than from the team, say the word and I'll send it through Gerrit against those branches — happy either way, I just didn't want to leave a stale conflicting PR sitting here.

Thanks for following up, and for pointing at the Gerrit change.

PR #1055, opened 2026-06-16
dns: Fix memory leak in dns_server_addr_parse

saddamr3e:
When parsing server addresses in dns_server_addr_parse(), if the DNS server's address count is already full (exceeds the limit of 8), the function returned early without freeing the successfully resolved addrinfo struct.
Fix this by moving the server->addr_count limit check above openvpn_getaddrinfo(), which avoids both the memory leak and the redundant DNS lookup.

### DIFF
--- a/src/openvpn/dns.c
+++ b/src/openvpn/dns.c
@@ -114,13 +114,13 @@ dns_server_addr_parse(struct dns_server *server, const char *addr)
         addr = addrcopy;
     }
 
-    struct addrinfo *ai = NULL;
-    if (openvpn_getaddrinfo(0, addr, NULL, 0, NULL, af, &ai) != 0)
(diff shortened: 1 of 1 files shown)


cron2:
This looks like a valid finding, thanks very much.

Since we do not deal with GH pull requests, we need to have it on the openvpn-devel@lists.sourceforge.net mailing list, or in gerrit.openvpn.net - you can do that, or for first-time submitters one of the regular contributors can do it (but it might take a bit of time).

cron2:
Patch has been pushed to http://gerrit.openvpn.net/c/openvpn/+/1712

saddamr3e:
> Patch has been pushed to http://gerrit.openvpn.net/c/openvpn/+/1712

thanks

PR #1091, opened 2026-08-05
metric: Enable support for route metric on FreeBSD

spmzt:
The route metric has been available since FreeBSD version 1600019.

URL: https://www.mail-archive.com/openvpn-devel@lists.sourceforge.net/msg38124.html


cron2:
https://www.mail-archive.com/openvpn-devel@lists.sourceforge.net/msg38124.html

Issue #1094, opened 2026-08-11
TCP short writes cause stream corruption


anluerz:
**Describe the bug**
In TCP mode, [`link_socket_write_tcp_posix()`](https://github.com/OpenVPN/openvpn/blob/da1b2a7081eb2791f3a675945c42d7f3189ef752/src/openvpn/socket.h#L717) calls `send()` exactly once with no retry loop. POSIX does not guarantee that `send()` on a stream socket writes all requested bytes, i.e. a partial write is a valid outcome. When this occurs, the unwritten tail of the packet is dropped and the TCP framing is permanently broken. Subsequent packets are misframed, decryption fails, and the session restarts.
The [caller in `forward.c`](https://github.com/OpenVPN/openvpn/blob/da1b2a7081eb2791f3a675945c42d7f3189ef752/src/openvpn/forward.c#L1809) [detects the size mismatch](https://github.com/OpenVPN/openvpn/blob/da1b2a7081eb2791f3a675945c42d7f3189ef752/src/openvpn/forward.c#L1830) and logs a warning, but takes no corrective action (no retry, no connection teardown). The stream is left in a corrupt state.

**To Reproduce**
We observed the following log sequence on the server side when the peer came under memory pressure:
```
vpn-seed-server 2026-07-15 20:02:01 vpn-shoot-client-1/tcp6-server:[REDACTED]:45670 Fatal decryption error (process_incoming_link), restarting
vpn-seed-server 2026-07-15 20:02:01 vpn-shoot-client-1/tcp6-server:[REDACTED]:45670 SIGUSR1[soft,decryption-error] received, client-instance restarting
vpn-seed-server 2026-07-15 20:02:01 Closing tun/tap interface
```
The behavior can also be reproduced deterministically with an LD_PRELOAD shim that intercepts send() and injects a short write. (PR with the reproducer about to follow)

**Expected behavior**
On a short write, send should be retried for the remaining bytes. The truncation path in forward.c should not leave the stream in an inconsistent state.

**Version information**
* OS: Alpine Linux
* OpenVPN version: 2.7.3
* Peer OS: Alpine Linux
* Peer OpenVPN version: 2.7.3

schwabe:
This is the first report that this can actually happen in decades. The probably requires quite extraordinary circumstances on the client.

anluerz:
[...]

PR #1095, opened 2026-08-11
Add TCP short-write regression test

anluerz:
This PR adds a self-contained regression test that demonstrates the bug reported in issue https://github.com/OpenVPN/openvpn/issues/1094: link_socket_write_tcp_posix() calls send() once with no retry, and a short write permanently corrupts the TCP framing.

The test currently FAILs (bug present). It is intended to be paired with a follow-up fix, after which it should PASS.

### What's included

- `tests/short_write_shim.c`:  An LD_PRELOAD shim that intercepts send() and injects a short write on P_DATA_V1/P_DATA_V2 packets (opcodes 6 and 9), while letting TLS handshake traffic pass through unmodified. This ensures the session is fully established before the data channel is corrupted.
-  `tests/t_short_write.sh`: A  test script that starts a loopback TCP server, runs a client with the shim preloaded, and checks whether the truncation warning appears in the log. Exit 1 (FAIL) = bug present; exit 0 (PASS) = bug fixed; exit 77 (SKIP) = prerequisites missing (e.g. macOS, missing binary).
- `tests/Makefile.am`: shim and script wired into the test suite under a TARGET_LINUX guard.
- `configure.ac`: adds the TARGET_LINUX conditional (mainly because I have no other platforms available)

### How to run

```sh
make check-TESTS TESTS=t_short_write.sh
```

No root or tun device required the test uses `--dev null --ping 1` with the existing sample keys.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -376,6 +376,7 @@ case "$host" in
 esac
 
 AM_CONDITIONAL([CROSS_COMPILING], test "${cross_compiling}" = "yes")
+AM_CONDITIONAL([TARGET_LINUX], [case $host in *-*-linux*) true;; *) false;; esac])
 
 PKG_PROG_PKG_CONFIG
 # Add variable to print if pkg-config is found or not. Users often miss that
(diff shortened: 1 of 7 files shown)


qjoner:
make check-TESTS TESTS=t_short_write.sh

PR #1101, opened 2026-09-07
buffer: return false from buf_puts() on truncation

cuishuang:

---

`buf_puts()` documents that it returns true when the string fits in the buffer and false when it is truncated.

However, the current implementation returns true whenever at least one byte of forward capacity remains. This is incorrect because `strncpynt()` may truncate the string to make room for the terminating NUL byte.

For example, with a 5-byte buffer:

```c
struct buffer buf = alloc_buf(5);
buf_puts(&buf, "12345");
```

the buffer contains `"1234"`, but `buf_puts()` currently returns true.

When only one byte of capacity remains, writing `"x"` stores an empty string and still returns true.

Measure the source string length before copying and return `len < cap`, where `cap` includes the terminating NUL byte.

This makes the return value match the documented API contract:

- `"1234"` in a 5-byte buffer returns true.
- `"12345"` in a 5-byte buffer returns false and stores `"1234"`.
- `"x"` in a 1-byte buffer returns false and stores an empty string.

### DIFF
--- a/src/openvpn/buffer.c
+++ b/src/openvpn/buffer.c
@@ -252,15 +252,16 @@ buf_printf(struct buffer *buf, const char *format, ...)
 bool
 buf_puts(struct buffer *buf, const char *str)
 {
-    int ret = false;
+    bool ret = false;
     uint8_t *ptr = BEND(buf);
     int cap = buf_forward_capacity(buf);
     if (cap > 0)
     {
+        const size_t len = strlen(str);
         strncpynt((char *)ptr, str, cap);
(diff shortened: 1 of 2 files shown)


cuishuang:
send

image

flichtenheld:
I'm actually not sure whether this is a bug in the code or whether the documentation I added was just wrong.

It is also meaningless since no users actually cares about the return value.

flichtenheld:
> send
> image

This mail did not make it to the list. I think you need to be subscribed first before you can send to the list.

